Issue #35, opened 2022-10-06
Add vignette 


joshwlambert:
A vignette that describes the functionality of `epiparameter`.

joshwlambert:
First version of the vignette is now live on the main branch and pkgdown site. Closing this issue for now.

PR #37, opened 2022-10-17
Feature/vignette

joshwlambert:
**New vignette explaining the use of `epiparameter`**

PR features:
- Added vignette to the package explaining the basic functionality of the package as well as cases in which the data stored within can be used in epidemiological analysis pipelines and interface with widely used packages (e.g. `EpiEstim` and `EpiNow2`). (#35)
- Added functions that convert between the parameters of the gamma and lognormal distribution and the mean and standard deviation of the distribution. This removes the dependency on `epitrix` for two functions and formulates the conversion using standard deviation instead of coefficient of variation as sd is more commonly used by other packages.
- Added plotting method for the `epidist` S3 class to easily facilitate the visualisation of the CDF, PDF and PMF.
- Added `pathogen_summary()` function that provides an easy to read output of the delay distributions available in `epiparameter` for a specific pathogen.

Discussion points:
- The plotting method for `epidist` uses `mtext()`([see here](https://github.com/epiverse-trace/epiparameter/blob/feature/vignette/R/epidist.R#L203)) to put text onto the facetted plot. This is not fixed in place and thus has formatting issues when the `day_range` argument is changed, see the last plot in the vignette for an example of how this goes wrong. Before merging the PR, I think we should try and find a solution that fixes the placement of the text. I am not opposed to just removing the text from this plotting function as it is not strictly necessary.
- Currently the idea for collaboration on the `epiparameter` library of delay distributions is to contribute to a google sheet that will then be updated periodically in the R package after being checked (this is a short term solution and is likely not going to be the permanent method of contribution). Please check the link to the google sheet in the vignette works as expected.
- There is a note on the RMD check because the data saved for `EpiNow2::epinow()` is fairly large and causes the size of the package to be 5.5Mb. This is a trade-off between run time for the vignette and the size of the data (and thus the package). Please let me know if there is a more optimal solution.
- The plotting method from `R0` was behaving weirdly and after a single plot was in the viewing window (of RStudio) the subsequent plots opened as separate windows, as well as this the plot was not rendering in the vignette (possibly for the same underlying reason). As a result I have written a plotting function that gives a very similar output to the plotting method of `R0` but is concealed in the vignette.


vignettes/epiparameter.Rmd
```diff
@@ -0,0 +1,521 @@
...
+library(EpiEstim)
+library(EpiNow2)
+library(R0)
+library(outbreaks)
+library(incidence2)
+```
+
+`epiparameter` is an R package that contains a range of delay distributions collated from peer-reviewed reports and meta-analyses, as well as functions to extract the parameters of a right-skewed parametric probability distribution from its percentiles or a median and range. If you would like to contribute to the database on delay distributions in this R package see the [contributing section](#contributing) at the end of this document.
+
+Definitions:
```
Bisaloo, line 33: This still needs to be completed, right?

vignettes/epiparameter.Rmd
```diff
@@ -0,0 +1,521 @@
...
+
+Delay distributions included in `epiparameter`:
+
+-   incubation period: The time delay between the time of infection and the onset of symptoms
+-   onset to admission: The time delay between the time of symptom onset to admission to hospital
+-   onset to death: The time delay between the time of illness onset to death
+-   serial interval: The time delay between the onset of symptoms between a primary case and a secondary case
```
Bisaloo, line 40: The list of implemented delay distributions should have a corresponding `match.arg()` in `list_distributions()` so we return a helpful error message if the user request something epiparameter doesn't provide.

vignettes/epiparameter.Rmd
```diff
@@ -0,0 +1,521 @@
...
+    y = c(rev(as.vector(out$estimates$TD$conf.int["lower"])$lower),
+          as.vector(out$estimates$TD$conf.int["upper"])$upper), 
+    col = "grey80", 
+    border = NA
+  )
+  
+  lines(as.Date(names(out$estimates$TD$R)), out$estimates$TD$R)
+}
+
+plot_R0(out)
```
Bisaloo, line 394: I think it will be confusing for users if we rely on this trick of showing a chunk of code and evaluating another.

What don't you like in the default plotting method?

vignettes/epiparameter.Rmd
```diff
@@ -0,0 +1,521 @@
...
+```
+
+Using the same method to calculate $R_t$ but this time using the data on the outbreak of ebola from Sierra Leone from 2014 and using the a serial interval from `epiparameter` we can again estimate $R_t$.
```
Bisaloo, line 397: You can remove the example with the `Germany.1918` example and jump directly to this section, which is much more relevant for potential epiparameter users.

vignettes/epiparameter.Rmd
```diff
@@ -0,0 +1,521 @@
...
+```{r, run R0 ebola, eval=FALSE}
+out <- estimate.R(
+  epid = ebola_incidence_R0, 
+  GT = mGT, 
+  begin = 1, 
+  end = 27, 
+  methods = "TD", 
+  nsim = 1000
+)
+```
```
Bisaloo, line 439: Is it too slow to run in the vignette? If so, I would still try to include it and reduce `nsim` value.

vignettes/epiparameter.Rmd
```diff
@@ -0,0 +1,521 @@
...
+    std_std_si = 0.5,
+    min_std_si = 0.5, 
+    max_std_si = 2.5
+  ))
+)
+```
+
+## Contributing to library of delay distributions {#contributing}
+
+If you would like to contribute to the different delay distributions stored in the `epiparameter` package, you can access the [google sheet](https://docs.google.com/spreadsheets/d/1VzC8NFHSQYPKsiDnxA7gDsm4cVr79outKQZJP8cC7Ro/edit?usp=sharing) and add your data.
```
Bisaloo, line 505: I can view the sheet but I have to request access if I want to edit it. Is that was you're going after?

vignettes/epiparameter.Rmd
```diff
@@ -0,0 +1,521 @@
...
+          as.vector(out$estimates$TD$conf.int["upper"])$upper), 
+    col = "grey80", 
+    border = NA
+  )
+  
+  lines(as.Date(names(out$estimates$TD$R)), out$estimates$TD$R)
+}
+
+plot_R0(out)
+```
```
Bisaloo, line 395:
````suggestion
```{r plot R0, fig.align = "center", fig.height = 8, fig.width = 8, out.width = "500px", fig.cap="Estimated value of $R_t$ for the incidence data of Spanish flu for Germany 1918-1919."}
plot(out$estimates$TD)
```
````

File: vignettes/epiparameter.Rmd [hunk 8]
```diff
@@ -0,0 +1,521 @@
...
+)
+```
+
+```{r plot R0 ebola, eval=FALSE}
+plot(out)
+```
+
+```{r plot Rt ebola, echo=FALSE, fig.align = "center", fig.height = 8, fig.width = 8, out.width = "500px", fig.cap="Estimated value of $R_t$ for the incidence data of ebola for Sierra Leone 2014."}
+plot_R0(out)
+```
```
Bisaloo, line 447:
````suggestion
```{r plot R0 ebola, fig.align = "center", fig.height = 8, fig.width = 8, out.width = "500px", fig.cap="Estimated value of $R_t$ for the incidence data of ebola for Sierra Leone 2014."}
plot(out$estimates$TD)
```
````

File: R/epidist.R [hunk 9]
```diff
@@ -137,3 +137,83 @@ print.epidist <- function(x, ...) {
 
   invisible(x)
 }
+
+#' @export
```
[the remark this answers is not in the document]
joshwlambert: This function has now been documented (2bbc04074c116b4a3f6494b5db17e0413a7b66c5)

R/pathogen_summary.R
```diff
@@ -0,0 +1,78 @@
...
+  
+  pathogen_id <- NULL # remove global variable note
+  
+  # Extract pathogen and parameter type
+  pathogen_data <- utils::read.csv(file = system.file(
+    "extdata",
+    "parameters.csv",
+    package = "epiparameter",
+    mustWork = TRUE
+  )) |> dplyr::filter(pathogen_id == pathogen)
```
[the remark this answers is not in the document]
joshwlambert: All uses of the pipe and dplyr have now been removed along with the dependency on a recent R version. (465eb920a2a94c61aa542d4c185e87582e0c2ba9)

vignettes/epiparameter.Rmd
joshwlambert: I wasn't planning on adding anything else to this section. Is there something in particular you think needs adding?

vignettes/epiparameter.Rmd
joshwlambert: Will add `match.arg()` to `list_distributions()` do you also think I should add `match.args()` to the other functions that take string arguments?

joshwlambert: Added `match.arg()` for functions that take character strings as arguments (3e65f383f54a10826b0d21ab9dbe731112fb3f1c)

Bisaloo:
@joshwlambert, I'm going to unsubscribe from this PR to avoid notifications while you're making changes. Please ping me when you'd like another round of review.

vignettes/epiparameter.Rmd
joshwlambert: Done (4e6f9a6ef06131289e4c913710db7af21c162486)

vignettes/epiparameter.Rmd
joshwlambert: Removed the saved data and loading it into the vignette. Instead have reduced `nsim` to 100 and explained that a higher number should be used in "real" applications (4e6f9a6ef06131289e4c913710db7af21c162486)

vignettes/epiparameter.Rmd
joshwlambert: I've updated the settings on the google sheet to enable anyone to edit. Slightly less safe but I think it will just be a lot easier than people having to request editing access. Thanks for pointing this out.

R/list_distributions.R
```diff
@@ -24,6 +26,14 @@
 
 list_distributions <- function(delay_dist = "incubation", parameters = FALSE) {
 
+  delay_dist <- match.arg(
+    arg = delay_dist, 
+    choices = c(
+      "incubation", "onset_to_admission", "onset_to_death", "serial_interval"
+    ),
```
Bisaloo, line 33: Could you put this directly in the function definition as you did in `extract_param()` please?

R/pathogen_summary.R
```diff
@@ -0,0 +1,92 @@
...
+pathogen_summary <- function(pathogen) {
+  
+  pathogen <- match.arg(
+    arg = pathogen, 
+    choices = c(
+      "adenovirus", "ebola", "human_CoV", "influenza_A_seasonal", 
+      "influenza_B_seasonal", "influenza_H1N1p", "influenza_H7N9", "marburg", 
+      "measles", "MERS_CoV", "parainfluenza", "rhinovirus", "RSV", "SARS_CoV", 
+      "SARS_CoV_2_wildtype", "monkeypox"
+    ),
```
Bisaloo, line 24: It seems like a lot of work to update this list each time new data is added to the spreadsheet. It might be a better strategy to generate this list dynamically.

For now, it's probably good enough to generate it each the function runs but in the future, we should probably figure out a way to generate it just once when the spreadsheet is updated.

R/pathogen_summary.R
```diff
@@ -0,0 +1,92 @@
...
+    choices = c(
+      "adenovirus", "ebola", "human_CoV", "influenza_A_seasonal", 
+      "influenza_B_seasonal", "influenza_H1N1p", "influenza_H7N9", "marburg", 
+      "measles", "MERS_CoV", "parainfluenza", "rhinovirus", "RSV", "SARS_CoV", 
+      "SARS_CoV_2_wildtype", "monkeypox"
+    ), 
+    several.ok = FALSE
+  )
+  
+  pathogen_id <- NULL # remove global variable note
```
Bisaloo, line 28: Not necessary now we got rid of dplyr

R/epidist.R
```diff
@@ -47,7 +47,26 @@
...
+  
+  delay_dist <- match.arg(
+    arg = delay_dist, 
+    choices = c(
+      "incubation", "onset_to_admission", "onset_to_death", "serial_interval"
+    ), 
+    several.ok = FALSE
+  )
+  
   pathogen_id <- NULL # remove global variable note # change to snake_case
```
Bisaloo, line 70: Not necessary now that we got rid of dplyr

vignettes/epiparameter.Rmd
```diff
@@ -0,0 +1,459 @@
...
+epidist(pathogen = "SARS_CoV", delay_dist = "incubation", study = "Donnelly_etal")
+```
+
+Now we have a distribution, in this case the lognormal distribution and its best-fit parameters for the incubation period. The distributions easily be visualised.
+
+```{r, plot epidist, fig.align = "center", fig.height = 8, fig.width = 8, out.width = "500px"}
+plot(sars_cov_incub)
+```
+
+In order to determine the duration of a quarantine period for SARS-CoV we can calculate the percentles of the distribution. For example, if we calculate the 95th percentile of the lognormal distribution for our SARS-CoV example it is the number of days within which 95% of individuals will have their incubation period elapsed.
```
Bisaloo, line 132:
```suggestion
In order to determine the duration of a quarantine period for SARS-CoV we can calculate the percentiles of the distribution. For example, if we calculate the 95th percentile of the lognormal distribution for our SARS-CoV example it is the number of days within which 95% of individuals will have their incubation period elapsed.
```

vignettes/epiparameter.Rmd
```diff
@@ -0,0 +1,459 @@
...
+names(ebola_incidence_R0) <- ebola_incidence$date_index
+
+ebola_si <- epidist(pathogen = "ebola", delay_dist = "serial_interval")
+
+si_mean_sd <- gamma_shapescale2meansd(
+  shape = ebola_si$param[["shape"]], 
+  scale = ebola_si$param[["scale"]]
+)
+
+mGT <- generation.time(type = "gamma", val = c(si_mean_sd$mean, si_mean_sd$sd))
```
Bisaloo, line 367: Would it be worth mentioning we take the serial time as a proxy for the generation time, which is not exact but probably good enough?
Maybe something to confirm with Adam.

tests/testthat/test-epidist.R
```diff
@@ -47,5 +47,21 @@ test_that("epidist works with lognorm dist", {
 })
 
 test_that("epidist.print works as expected", {
-  expect_snapshot(epidist(pathogen = "RSV", delay_dist = "incubation"))
+   expect_snapshot(epidist(pathogen = "RSV", delay_dist = "incubation"))
 })
+
+test_that("epidist.plot does not produce an error", {
+  expect_silent(plot(epidist(pathogen = "ebola", delay_dist = "incubation")))
```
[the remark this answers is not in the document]
joshwlambert: Implemented in f56688b8d104802545b5c4be9838ac6fdf41a0e6

R/pathogen_summary.R
```diff
@@ -0,0 +1,78 @@
...
+  )
+  colnames(pathogen_tbl) <- c(
+    "delay_dist", "distribution", "mean", "sd", "study", "DOI"
+  )
+  
+  out <- list(
+    pathogen = pathogen,
+    pathogen_tbl = pathogen_tbl
+  )
+
```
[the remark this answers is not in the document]
joshwlambert: Updated in 2ad960f8a60de420ba3f65b453601f5d6fa3648e

R/pathogen_summary.R
```diff
@@ -0,0 +1,78 @@
...
+        scale = as.numeric(x[["scale"]])
+      )
+    } else if (x[["distribution"]] == "lnorm") {
+      parameters <- lnorm_musigma2meansd(
+        mu = as.numeric(x[["meanlog"]]), 
+        sigma = as.numeric(x[["sdlog"]])
+      )
+    }
+    parameters
+  })
```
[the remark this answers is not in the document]
joshwlambert: I will leave this function for now as I think the future use of a dedicated distributions package will render this redundant, but I will remove once that change has been made.

R/pathogen_summary.R
joshwlambert: I agree, especially as the number of pathogens grows. I'll create an issue for this to be changed.

R/list_distributions.R
joshwlambert: Done bc4b5db42937115d74bc928b5082736a7a29f3a1

tests/testthat/test-pathogen_summary.R
```diff
@@ -0,0 +1,37 @@
+test_that("pathogen_summary works with valid input", {
+  ebola_summary <- pathogen_summary(pathogen = "ebola")
+  expect_true(is.data.frame(ebola_summary))
```
Bisaloo, line 3:
```suggestion
  expect_s3_class(ebola_summary, "data.frame")
```

tests/testthat/test-pathogen_summary.R
```diff
@@ -0,0 +1,37 @@
...
+  expect_equal(nrow(ebola_summary), 3)
+  expect_equal(ncol(ebola_summary), 7)
+  expect_equal(
+    colnames(ebola_summary),
+    c("pathogen", "delay_dist", "distribution", "mean", "sd", "study", "DOI")
+  )
```
Bisaloo, line 9:
```suggestion
  expect_named(
    ebola_summary,
    c("pathogen", "delay_dist", "distribution", "mean", "sd", "study", "DOI")
  )
```

tests/testthat/test-pathogen_summary.R
```diff
@@ -0,0 +1,37 @@
...
+  expect_true(is.character(ebola_summary$pathogen))
```
Bisaloo, line 10:
```suggestion
  expect_type(ebola_summary$pathogen, "character")
```

Bisaloo (approved):

vignettes/epiparameter.Rmd
```diff
@@ -0,0 +1,465 @@
...
+reported_cases <- example_confirmed[1:40]
+
+# use lnorm_logmeansd2meansd() to convert the parameters of the lognormal to the
+# mean (expectation) and standard deviation
+serial_interval_mean_sd <- lnorm_logmeansd2meansd(
+  meanlog = serial_interval$param[["meanlog"]],
+  sdlog = serial_interval$param[["sdlog"]]
+)
+
+incubation_period_mean_sd <- lnorm_logmeansd2meansd(
```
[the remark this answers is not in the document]
joshwlambert: This was an old function that was renamed. Good evidence of why `eval=FALSE` blocks are bad. I've move this code into an evaluated chunk and only the `epinow()` function is not evaluated.

PR #49, opened 2022-11-01
Feature/conv weibull

joshwlambert:
The main objective of this pull request is to add the functionality to convert between the parameters of the Weibull distribution and its mean and standard deviation. Additional changes include:
- updating the linting workflow to highlight linting in files that have changed
- change the default of `list_distributions()` to not filter by incubation period, but to return all delay distributions when no argument is provided (or `"all"`) is provided
- improved testing expectations by checking for specific types
- general package linting


R/list_distributions.R
```diff
@@ -41,26 +43,30 @@ list_distributions <- function(
...
 
-  if (parameters == TRUE) {
-    output <- params
+  if (delay_dist != "all") {
+    # filter by delay distribution
+    params <- params[params$type_id == delay_dist, ]
   }
-  if (parameters == FALSE) {
+
+  if (isFALSE(parameters)) {
```
Bisaloo, line 60: Any reason to prefer `isFALSE(parameters)` to `!parameters` in this specific case?

If you want to keep `isFALSE()`, you'll need to add a dependency on R 3.5.0 in `DESCRIPTION`

Bisaloo (approved):

R/convert_param.R
```diff
@@ -67,3 +67,47 @@ gamma_meansd2shapescale <- function(mean, sd) {
...
+#'
+#' @return A named list with shape and scale parameters
+#' @export
+#'
+#' @examples
+#' weibull_meansd2shapescale(mean = 1.5, sd = 0.5)
+weibull_meansd2shapescale <- function(mean, sd) {
+
+  # give warning message about numerial inaccuracies
+  message("Numerical approximation used, results may be unreliable.")
```
Bisaloo, line 85: Would it be helpful to give users a sense of how precise / rough this approximation is?

Maybe something like this in the examples?

```r
wss <- weibull_meansd2shapescale(mean = 1.5, sd = 0.5)
weibull_shapescale2meansd(wss[["shape"]], wss[["scale"]])
```

R/list_distributions.R
joshwlambert: I thought it was safer, in cases where `parameters` might be a vector. I'll add the dependency to R 3.5.0. This will also cover cases where `WARNING: Added dependency on R >= 3.5.0 because serialized objects in
     serialize/load version 3 cannot be read in older versions of R.` appear in the check because of using `.rda` data for the vignette.

joshwlambert: Resolved in 6ec8ad7fce350054bb13fe75453bf3924b1fdbd4

R/convert_param.R
joshwlambert: Resolved in 4e6213f8a4298d6cc8aeb74d206befe75e769d8c

PR #84, opened 2022-12-21
Add google analytics

Bisaloo:
I'm targeting `main` on purpose because we likely want this deployed as soon as possible, and not wait until `develop` is ready. `develop` is not likely to touch `_pkgdown.yml` so I don't anticipate major conflicts.


joshwlambert (approved):
Everything looks good. Thanks for making the update.

Issue #83, opened 2022-12-16
Add pdf, cdf, quantiles and random sampling methods for epidist class


joshwlambert:
The `epidist` class stores a probability distribution, either as a continuous distribution from the {`distributional`} package or a discrete distribution from the {`distcrete`} package. Instead of requiring the user to access the slot containing the distribution and calling methods from either of the aforementioned packages I propose writing methods for the `epidist` class that will call the methods of each package. 

The additional benefit of this is {`distributional`} and {`distcrete`} have different function names for quantiles, densities, random number generation, etc. By writing a wrapper the user will get a consistent interface and will not have to worry about whether there `epidist` object holds a continuous or discrete distribution.

joshwlambert:
I propose following the {`distributional`} naming convention for these functions as they are more self-explanatory:
```
density()
cdf()
quantile()
generate()
```
As opposed to the {`distcrete`} style:
```
d()
p()
q()
r()
```

Happy to hear suggestions for other names or changes to the alternative style (as this will be familiar to those using base R distributions).

Issue #82, opened 2022-12-16
Add right-truncation option for continuous distributions


joshwlambert:
Distributions can be right-truncated to adjust for delays that have a primary event but not a secondary event. The distribution objects from {`distributional`} can have left or right truncation applied. This should be integrated into the `epidist` class to handle right-truncated distributions.

Issue #81, opened 2022-12-15
Allow discrete distributions using {distcrete}


joshwlambert:
Currently the new `epidist` class uses continuous distribution objects from {`distributional`} but this package does not contain discretised versions of gamma, lognormal and weibull distributions. Therefore, depending on {`distcrete`} will allow these to be used. This relates to issue #6.

joshwlambert:
This feature was implemented in PR #85, and therefore this issue is being closed.

Issue #74, opened 2022-12-02
Add epiparam class


joshwlambert:
The data that is stored in the `epiparameter` package is loaded and manipulated and viewed through functions like `list_distributions()` and `pathogen_summary()`, the `epiparam` class will standardise the data and how functions interact with it (by becoming class methods).

Related to issue #48

PR #85, opened 2023-01-31
New package architecture 

joshwlambert:
This pull request contains a large number of changes influencing the handling of epidemiological data, the API of the `epidist` class and its methods, as well as work on numerous helper functions. 

Changes to the `{epiparameter}` library:
- The library has several new categories
    - both disease and pathogen in library #63 
    - to account for the various ways in which parameters, summary statistics and uncertainty can be reported. 
    - There are columns for methodological criteria used when inferring the distribution (e.g. censored, right-truncated).

Changes to the epidist class:
- The epidist class now does not directly read from the `{epiparameter}` library. Instead, it is a stand alone class that can be constructed by the user. 
- To keep the useful functionality of the old `epidist()` function of reading directly from the library, the `epidist_db()` function has been added.
- The "encapsulated" functions that were part of the previous version of the `epidist` class have been replaced with class methods. These have new names that follow from the naming convention of the `{distributional}` package (which is depended on to handle distributions). #73 
- The `epidist` class can handle continuous (with right-truncation #82) and discretised distributions (using `{distributional}` or `{distcrete}`). #81, #83 
- The `epidist` class now follows the structure of having a constructor, helper and validator function.

Addition of an `epiparam` class:
- In the old code the epiparameter library could be read in and viewed through several functions (`epidist()`, `list_distributions()` and `pathogen_summary()`), but there was no way to handle the library and manipulate it in custom ways. The `epiparam` class is a data structure for the library and is a subclass of a `data.frame` to provide printing and subsetting methods. #74 
- The `epiparam` class follows the structure of having a constructor, helper and validator function.

Addition of a `vb_epidist` class:
- In order to maintain all the functionality of the `epidist` class for vector-borne diseases in which there is an intrinsic and extrinsic delay (e.g. incubation period) the `vb_epidist` class provides a consistent interface by wrapping around the `epidist` class. #50 
- The `vb_epidist` class has the same probability distribution methods as the `epidist` class.
- The `vb_epidist` class follows the structure of having a constructor, helper and validator function.

Addition of JSON validation on library:
- There is a new github action workflow to validate the epiparameter data.
- There are helper functions in utils.R to help update the JSON data and JSON schema from the data dictionary yaml file and library csv file. #64 

Addition of conversion functions:
- Negative binomial
- Geometric

There are class conversion functions (`as_epidist()` and `as_epiparam()`). There is also a `bind_epiparam()` function that allows all classes defined in the `{epiparameter}` package to be bound to an existing `epiparam` object.

Helper functions for constructing objects have been added to aid in supplying information manually when constructing an `epidist` object. For example `create_epidist_citation()` and `create_epidist_metadata()`.

Two exported functions, `calc_dist_params()` and `create_prob_dist()` are added which are mainly used internally but are exported in case users may find them helpful.

Added data protocol vignette that describes the process of building the `{epiparameter}` library. #68 

The `pathogen_summary()` function has been removed as it's functionality was no longer deemed necessary. 

Documentation and testing have been updated in line with the above-mentioned developments.


.github/workflows/validate-json.yaml
```diff
@@ -0,0 +1,39 @@
...
+        run: |
+          jsonvalidate::json_validate(
+            json = jsonlite::read_json(path = "inst/extdata/data.json")[[1]],
+            schema = jsonlite::read_json(path = "inst/extdata/schema.json")[[1]],
+            engine = "ajv",
+            verbose = TRUE,
+            greedy = TRUE,
+            strict = TRUE,
+            error = TRUE
+          )
```
pratikunterwegs, line 38: I've spent a lot of time trying to figure out why this workflow fails as it does, and what alternatives there are. Can work on this further as well with a simpler example, but for now a few points:

1. The workflow fails as-is becasue some elements do not have the correct values for `prob_distribution`, these are rows 26:29, but checking those, it seems the values are more or less correct ("weibull" and NA), so unclear what's happening. The ones which are NA might be failing as the `prob_distr` field isn't present in the JSON.
2. When saving these in 'pretty' format and running the validator, it fails on checking whether the database is a JSON object; the error is not at all informative.

`{jsonvalidate}` requires a character string (or path) for both `json` and `schema`, and it's possible to get that by adding `jsonlite::toJSON()` around `jsonlite::read_json()`. That allows you to save the files as human-readable JSON, rather than the unreadable string currently used.

pratikunterwegs (commented):
Overall this is a major change to `{epiparameter}`, and overall it looks good. I'm writing my thoughts and a few suggestions here for now and will add file-specific comments later. We should probably avoid such large PRs generally as they do pose a bit of a challenge. Feel free to implement suggestions as appropriate - I think the overall functionality is there, and polishing fixes could be (should be!?) pushed to a later PR.

Conceptually, making a distinction between `epidist` objects and the parameter library (accessed through `epiparam`) is probably a good step. I think the additions to the `epidist` class make it quite comprehensive, while also not requiring all the fields to be populated, and also including several helper functions to help fill those fields if needed.

Thinking ahead, I imagine that `{epiparameter}` could be something like `{renv}` for pathogen parameters - a useful way to share them and to ensure that distributed workflows are working with the same parameters. This could also be locally when working across many projects on the same disease, so that multiple `epidist`s to be converted into an `epiparam` and vice versa, and to be saved to file for later use, sharing, or version control.

One thing I'm unsure of is whether `epidist_db()` is necessary, as the `as_epidist()` function already allows conversion from an `epiparam` to `epidist`(s). Perhaps a function that does the opposite conversion for a list of `epidist`s would be useful, an extension of `bind_epiparam`? 

I also wonder whether the parameter library is going to hold data for vector borne diseases as well - currently it seems that the extrinsic distributions have to be modelled *ad hoc*. I don't think an `as_vb_epidist()` method needs to be implemented now, but something to think about going forward.

I spent a lot of time trying to figure out why the JSON validation workflow is failing, and looking into the database import/export as JSON. I think it's good to have this functionality, but there seems to be a mixing up of functions (see below).

Technically, I think everything else is about as quick as it could be. A `{styler}` run should help clean up the code, as a linter workflow is currently failing. Adding an updated Readme and a new or updated "Get started" vignette would be useful, but given the number of commits in this PR, could be pushed to a new PR focusing on long-form documentation alone.

# epidist
## epidist class
1. Print method:
	1. Print pathogen name as well, may be useful when a broad category of diseases (e.g. HPAI, EVD) is caused by multiple related pathogens (or strains).
	2. On pathogen name: possibly need a way to indicate when a pathogen is truly unknown (e.g. a limited 'mystery' outbreak), versus when it is not included in the database. In the first case, there could be different sources for the delay distributions (e.g. from initial field epi study), and for the pathogen identity (e.g. later sequencing study) - unsure how to handle these possibly niche cases. Overall, raises questions about the taxonomic resolution of pathogen names - stick with what's reported in the papers?
	3. Print study name as well; helps users check they have the right study when comparing between outbreaks.
	4. Avoid $\gamma$ symbol for gamma distributions - looks cool but may be unnecessarily confusing to new users (esp. as monospace version looks different from usual gamma symbol). Also suggest avoiding `lN` for lognormal - same reasoning although I appreciate the compact representation.
2. Plot method:
	1. Consider `type = "b"` with a line joining the points, easier to read than points alone.
	2. Consider different plot type (column?) for discretised distributions.
3. Summary statistics: Unclear from the description what this refers to - are these the summary statistics of the data reported in the study, and how do these summary stats relate to the distribution parameters? A working example in `create_epidist_summary_stats()` documentation would be very useful, could also be copied here. Overall suggest explaining this better.
4. Automatic parameter calculation: Suggest moving this argument up to just after the `summary_stats` argument as they are related. Also sugget adding some information as to how the parameters are estimated - is this a fitting step or simply conversion? See below under **epidist_db** that this might be slowing down creation from database - suggest informative message to indicate (potentially slow) fitting if such.

## create_epidist_uncertainty
1. Suggest renaming list elements to "ci" for the actual interval (e.g 95%, currently called "ci_interval"), renaming "ci_interval" (confidence interval interval?) to "ci_limits", a two element vector for the lower and upper limits (similar to `range()` order, smaller first); "ci_type" seems fine.
2. Suggest adding example for creating uncertainty around two parameters, as a number of distributions have two parameters - it's unclear how this should be done, or whether `epidist` objects only support uncertainty in one parameter.
3. Example for unkown uncertainty has superfluous `NA`, and does not match usage in `epidist` documentation examples - suggest removing `NA`.
4. Suggest renaming function title to "Specify distribution parameter uncertainty", current title is too long and is identical to the description.

## create_epidist_summary_stats
1. Suggest changing function title to not be the same as the description, suggest instead "Create epidist summary statistics".
2. Suggest renaming all "statistic_ci" to follow naming of `create_epidist_uncertainty()`: "ci" for the interval width, and "ci_limits" for the lower and upper limits.
3. Suggest adding a realistic working example.

## create_epidist_metadata
1. Function title needs fixing.
2. Suggest better explaining `extrinsic()` as only applicable for vector borne diseases - suggest moving details to Details, where it would be good to explain what an extrinsic delay distribution is.
3. Suggest optionally better handling of `region` argument as a named list of "country", "province", "location", and possibly "other" for extra information. This functionality might help future review/meta-analysis studies which might want to count studies in each region/country, or to map locations of study sites from which parameters are available. This is something initiatives are increasingly interested in, see e.g. https://www.viralemergence.org/blog/building-a-global-atlas-of-wildlife-disease-data.
4. Value field requires expansion - update the `@return`  field.
5. Examples field should be styled - run `styler::style_pkg()`.

## create_epidist_method_assessment
1. Fix function title;
2. Typo in "censorred";
3. Value field should be expanded - typing this out gets repetitive and boring but better complete than vague.

## epidist_db
1. Overall not convinced that this 'extra' way to access the parameter library is required - is `as_epidist()` not sufficient?
2. Suggest function title is an imperative rather than description ("Create ..." rather than "Creates ...").
3. Function is difficult to use, with error messages non-informative. For example, when looking for the onset to death distribution for EVD using `epidist_db(disease = "ebola virus disease", epi_dist = "onset_to_death")`, I get this error: `Error in match.arg(arg = disease, choices = unique(eparam$disease), several.ok = FALSE) : 'arg' should be “marburg virus disease”`, which could be replaced with the more informative "X distribution not availabe for Y disease".
4. Related to (2.) above, has the onset to death distribution for EVD been removed? Why? Note that this will break some workflows for `{datadelay}` (together with removal of encapsulated functions).
5. Trying to create an `epidist` object from the database takes a relativel long time (15-ish seconds) for some diseases (such as dengue; try: `dengue_incubation = epidist_db(disease = "dengue", epi_dist = "incubation_period")`) - I assume this is due to parameters being estimated from summary statistics - suggest printing a message to notify users that this may take a few seconds, otherwise it gives the (wrong) impression that something is broken.

## as_epidist
1. See comment (5) on speed and informative message above.

## epidist_distribution_functions
1. Suggest adding examples to the functions grouped under this heading, including `density`, `cdf` etc; especially adding separate examples for the `epidist` and `vb_epidist` classes.

# vb_epidist
## vb_epidist class
1. Typo in function title.
2. Expand on what the `intrinsic_epidist` and `extrinsic_epidist` objects refer to, and especially which one refers to the vector. Suggest renaming these to "human_epidist" and "vector_epidist" for better understanding.
3. Plot and print methods: See feedback on `epidist` objects, same applies here. Suggest using `{ggplot2}` to avoid difficult placement of plot titles. Adds a dependency but possibly useful in the long run, especially if seeking to also add uncertainty visualisations. Not a massive priority.

# epiparam
## epiparam class
1. Suggest expanding description to explain what the `epiparam` class is about and what a typical use case is.
	1. Am I right in assuming that it is primarily a way to access the data stored in the package? 
	2. Should it be possible to create a local and temporary working library of delay distributions, i.e., make a number of `epidist` objects and build an `epiparam` library from them? This could be useful functionality that adds to `bind_epiparam` for users building their own parameter libraries, for example, to prepare for contributing to `{epiparameter}`.
2. On the whole I liked the ability to make `epidist` objects from `epiparam` dataframes - however, the function `as_epidist()` was difficult to find without a vignette or Readme. Before I found it, I essentially suggested adding a similar function. I had also sketched out a workflow for these conversions, which I'm leaving here anyway in case it's useful, say for a vignette: *(a)* Read in the parameter library using `epiparam()`, *(b)* Subset - while reading or after - the library for a disease and a distribution, *(c)* Make `epidist` objects from each row of the `epiparam`, and potentially *(d)* build a workflow that takes these `epidists`. For example, when dealing with a new outbreak, get all onset to death distributions for that disease-causing pathogen from the library, make each an `epidist`, and run separate CFR calculations using each one - this basically gets at the question: what is the CFR if this new outbreak is caused by pathogen similar to one of several previously known ones?
3. Is there a reason that some disease name are capitalised while others are not? See especially diseases with proper nouns, such as "west nile fever" or "ebola virus disease" vs. "Hantavirus pulmonary syndrome" or "Rift Valley fever". See also naming of place-virus disease names, such as "Hantavirus" (no space) vs. "ebola virus" (with space).

## make_json_data and make_json_schema
1. I think `make_json_data()` is a useful overall, but I would restructure it to avoid using `jsonlite::validate()` and `jsonlite::toJSON()` - I don't think either of them are necessary. It's possible to pass the dataframe to `jsonlite::write_json()` directly, and I would suggest passing the option `pretty = TRUE` to write out a human-readable JSON file. As I understand it, the use of `toJSON` with `write_json` is not correct, see https://stackoverflow.com/questions/69617597/importing-editing-and-saving-json-in-a-simple-way.
2. `make_json_schema` and the saving of the data dictionary as a YAML file, which is then converted to JSON only for the purposes of validating the database, seems a bit convoluted. Is this a common workflow? I would suggest saving the JSON schema as a human-readable file as well, this makes it much easier to figure out changes when reviewing. That requires passing `pretty = TRUE` as well as `auto_unbox = TRUE` to get the `$schema` string.
3. Overall I'm a little wary of the `{jsonvalidate}` package - seems clunky and not particularly informative when it errors.

# General technical comments
1. Suggest using `checkmate::assert_string()` instead of `assert_character(..., len = 1)` - just a bit cleaner (e.g. in `calc_dist_params()`, `new_epiparam()`).
2. Suggest adding limits to all `checkmate::assert_number()` to avoid zero or negative value, esp. with sample sizes.
3. Suggest standardising all messages, warnings and errors to sentence case.
4. Suggest avoiding complex conditionals such as `if (!all(is.na(...)))` in favour of `any()` - this may require changing the order of control flow.
5. Suggest a `{styler}` run.
6. Write separate titles and descriptions for all functions.
7. Suggest writing an `is.epidist` method to avoid repeated calls to `if (inherits(obj, "epidist"))` - consider also for `epiparam`.
8. Suggest single quotes or backticks around all class names in messages to highlight that they are not variables in the global environment.
9. Suggest increasing test coverage on "utils.R".
10. Update the DESCRIPTION for author information.

R/epidist_utils.R
```diff
@@ -0,0 +1,605 @@
...
+      all(ci_type %in% c("confidence interval", "credible interval"))
+  )
+
+  # return list of parameter uncertainty
+  list(
+    ci = ci,
+    ci_interval = ci_interval,
+    ci_type = ci_type
+  )
+}
```
pratikunterwegs, line 47: 1. Suggest renaming list elements to "ci" for the actual interval (e.g 95%, currently called "ci_interval"), renaming "ci_interval" (confidence interval interval?) to "ci_limits", a two element vector for the lower and upper limits (similar to `range()` order, smaller first); "ci_type" seems fine.
2. Suggest adding example for creating uncertainty around two parameters, as a number of distributions have two parameters - it's unclear how this should be done, or whether `epidist` objects only support uncertainty in one parameter.
3. Example for unkown uncertainty has superfluous `NA`, and does not match usage in `epidist` documentation examples - suggest removing `NA`.
4. Suggest renaming function title to "Specify distribution parameter uncertainty", current title is too long and is identical to the description.

R/epidist_utils.R
```diff
@@ -0,0 +1,605 @@
...
+  # return list of metadata
+  list(
+    sample_size = sample_size,
+    region = region,
+    vector_borne = vector_borne,
+    vector = vector,
+    extrinsic = extrinsic,
+    inference_method = inference_method
+  )
+}
```
pratikunterwegs, line 135: 1. Function title needs fixing.
2. Suggest better explaining `extrinsic()` as only applicable for vector borne diseases - suggest moving details to Details, where it would be good to explain what an extrinsic delay distribution is.
3. Suggest optionally better handling of `region` argument as a named list of "country", "province", "location", and possibly "other" for extra information. This functionality might help future review/meta-analysis studies which might want to count studies in each region/country, or to map locations of study sites from which parameters are available. This is something initiatives are increasingly interested in, see e.g. https://www.viralemergence.org/blog/building-a-global-atlas-of-wildlife-disease-data.
4. Value field requires expansion - update the `@return`  field.
5. Examples field should be styled - run `styler::style_pkg()`.

R/epidist_utils.R
```diff
@@ -0,0 +1,605 @@
...
+  checkmate::assert_logical(right_truncated, len = 1)
+  checkmate::assert_logical(phase_bias_adjusted, len = 1)
+
+  # return method assessment list
+  list(
+    censorred = censorred,
+    right_truncated = right_truncated,
+    phase_bias_adjusted = phase_bias_adjusted
+  )
+}
```
pratikunterwegs, line 367: 1. Fix function title;
2. Typo in "censorred";
3. Value field should be expanded - typing this out gets repetitive and boring but better complete than vague.

R/epidist_utils.R
```diff
@@ -0,0 +1,605 @@
...
+      q_975 = q_975
+    ),
+    range = list(lower_range = lower_range, upper_range = upper_range),
+    dispersion = list(
+      dispersion = dispersion,
+      dispersion_ci = dispersion_ci,
+      dispersion_ci_interval = dispersion_ci_interval
+    )
+  )
+}
```
pratikunterwegs, line 257: 1. Suggest changing function title to not be the same as the description, suggest instead "Create epidist summary statistics".
2. Suggest renaming all "statistic_ci" to follow naming of `create_epidist_uncertainty()`: "ci" for the interval width, and "ci_limits" for the lower and upper limits.
3. Suggest adding a realistic working example.

R/epidist.R
```diff
@@ -1,242 +1,609 @@
...
+    truncation = truncation,
+    notes = notes
+  )
+
+  # call epidist validator
+  validate_epidist(epidist = epidist)
+
+  # return epidist object
+  epidist
+}
```
pratikunterwegs, line 303: Suggestions for **epidist** methods
1. Print method:
	1. Print pathogen name as well, may be useful when a broad category of diseases (e.g. HPAI, EVD) is caused by multiple related pathogens (or strains).
	2. On pathogen name: possibly need a way to indicate when a pathogen is truly unknown (e.g. a limited 'mystery' outbreak), versus when it is not included in the database. In the first case, there could be different sources for the delay distributions (e.g. from initial field epi study), and for the pathogen identity (e.g. later sequencing study) - unsure how to handle these possibly niche cases. Overall, raises questions about the taxonomic resolution of pathogen names - stick with what's reported in the papers?
	3. Print study name as well; helps users check they have the right study when comparing between outbreaks.
	4. Avoid $\gamma$ symbol for gamma distributions - looks cool but may be unnecessarily confusing to new users (esp. as monospace version looks different from usual gamma symbol). Also suggest avoiding `lN` for lognormal - same reasoning although I appreciate the compact representation.
2. Plot method:
	1. Consider `type = "b"` with a line joining the points, easier to read than points alone.
	2. Consider different plot type (column?) for discretised distributions.
3. Summary statistics: Unclear from the description what this refers to - are these the summary statistics of the data reported in the study, and how do these summary stats relate to the distribution parameters? A working example in `create_epidist_summary_stats()` documentation would be very useful, could also be copied here. Overall suggest explaining this better.
4. Automatic parameter calculation: Suggest moving this argument up to just after the `summary_stats` argument as they are related. Also sugget adding some information as to how the parameters are estimated - is this a fitting step or simply conversion? See below under **epidist_db** that this might be slowing down creation from database - suggest informative message to indicate (potentially slow) fitting if such.

R/epidist_db.R
```diff
@@ -0,0 +1,72 @@
...
+    Please report an issue with duplicated studies.")
+    eparam <- eparam[1, ]
+  }
+
+  # convert epiparam to epidist
+  edist <- as_epidist(x = eparam)
+
+  # return epidist (or list of epidist)
+  edist
+}
```
pratikunterwegs, line 72: 1. Overall not convinced that this 'extra' way to access the parameter library is required - is `as_epidist()` not sufficient?
2. Suggest function title is an imperative rather than description ("Create ..." rather than "Creates ...").
3. Function is difficult to use, with error messages non-informative. For example, when looking for the onset to death distribution for EVD using `epidist_db(disease = "ebola virus disease", epi_dist = "onset_to_death")`, I get this error: `Error in match.arg(arg = disease, choices = unique(eparam$disease), several.ok = FALSE) : 'arg' should be “marburg virus disease”`, which could be replaced with the more informative "X distribution not availabe for Y disease".
4. Related to (2.) above, has the onset to death distribution for EVD been removed? Why? Note that this will break some workflows for `{datadelay}` (together with removal of encapsulated functions).
5. Trying to create an `epidist` object from the database takes a relativel long time (15-ish seconds) for some diseases (such as dengue; try: `dengue_incubation = epidist_db(disease = "dengue", epi_dist = "incubation_period")`) - I assume this is due to parameters being estimated from summary statistics - suggest printing a message to notify users that this may take a few seconds, otherwise it gives the (wrong) impression that something is broken.

R/epiparam_utils.R
```diff
@@ -0,0 +1,447 @@
...
+
+  # remove names of list
+  names(out) <- NULL
+
+  # unpack from list if only one epidist
+  if (length(out) == 1) out <- out[[1]]
+
+  # return list of epidist objects
+  out
+}
```
pratikunterwegs, line 37: See comment on speed and informative message on `epidist_db()`

R/epidist.R
```diff
@@ -1,242 +1,609 @@
...
+  unlist <- length(x$prob_dist) == 1
+  if (inherits(x$prob_dist, "distcrete")) {
+    out <- x$prob_dist$q(p)
+  } else {
+    out <- stats::quantile(x$prob_dist, p = p)
+  }
+  out <- if (unlist) unlist(out, recursive = FALSE) else out
+  out
+}
+
```
pratikunterwegs, line 589: Suggest adding examples to the functions grouped under this heading, including `density`, `cdf` etc; especially adding separate examples for the `epidist` and `vb_epidist` classes.

R/vb_epidist.R
```diff
@@ -0,0 +1,309 @@
...
+    intrinsic_epidist = intrinsic_epidist,
+    extrinsic_epidist = extrinsic_epidist
+  )
+
+  # call epidist validator
+  validate_vb_epidist(vb_epidist)
+
+  # return vb_epidist object
+  vb_epidist
+}
```
pratikunterwegs, line 90: 1. Typo in function title.
2. Expand on what the `intrinsic_epidist` and `extrinsic_epidist` objects refer to, and especially which one refers to the vector. Suggest renaming these to "human_epidist" and "vector_epidist" for better understanding.
3. Plot and print methods: See feedback on `epidist` objects, same applies here. Suggest using `{ggplot2}` to avoid difficult placement of plot titles. Adds a dependency but possibly useful in the long run, especially if seeking to also add uncertainty visualisations. Not a massive priority.

R/epiparam.R
```diff
@@ -0,0 +1,263 @@
...
+    "right_truncated needs to be a logical" =
+      is.logical(epiparam$right_truncated),
+    "phase_biase_adjusted needs to be a logical" =
+      is.logical(epiparam$phase_bias_adjusted),
+    "DOI needs to be a character" =
+      is.character(epiparam$DOI)
+  )
+
+  invisible(epiparam)
+}
```
pratikunterwegs, line 140: 1. Suggest expanding description to explain what the `epiparam` class is about and what a typical use case is.
	1. Am I right in assuming that it is primarily a way to access the data stored in the package? 
	2. Should it be possible to create a local and temporary working library of delay distributions, i.e., make a number of `epidist` objects and build an `epiparam` library from them? This could be useful functionality that adds to `bind_epiparam` for users building their own parameter libraries, for example, to prepare for contributing to `{epiparameter}`.
2. On the whole I liked the ability to make `epidist` objects from `epiparam` dataframes - however, the function `as_epidist()` was difficult to find without a vignette or Readme. Before I found it, I essentially suggested adding a similar function. I had also sketched out a workflow for these conversions, which I'm leaving here anyway in case it's useful, say for a vignette: *(a)* Read in the parameter library using `epiparam()`, *(b)* Subset - while reading or after - the library for a disease and a distribution, *(c)* Make `epidist` objects from each row of the `epiparam`, and potentially *(d)* build a workflow that takes these `epidists`. For example, when dealing with a new outbreak, get all onset to death distributions for that disease-causing pathogen from the library, make each an `epidist`, and run separate CFR calculations using each one - this basically gets at the question: what is the CFR if this new outbreak is caused by pathogen similar to one of several previously known ones?
3. Is there a reason that some disease name are capitalised while others are not? See especially diseases with proper nouns, such as "west nile fever" or "ebola virus disease" vs. "Hantavirus pulmonary syndrome" or "Rift Valley fever". See also naming of place-virus disease names, such as "Hantavirus" (no space) vs. "ebola virus" (with space).

R/utils.R
```diff
@@ -36,3 +36,85 @@ calc_disc_dist_quantile <- function(prob, days, quantile) {
...
+  # convert data to json
+  data_json <- jsonlite::toJSON(data, dataframe = "columns", na = "null")
+  jsonlite::validate(data_json)
+
+  # save json data
+  jsonlite::write_json(data_json, path = write_path)
+
+  # invisibly return json data
+  invisible(data_json)
+}
```
pratikunterwegs, line 84: I think `make_json_data()` is a useful overall, but I would restructure it to avoid using `jsonlite::validate()` and `jsonlite::toJSON()` - I don't think either of them are necessary. It's possible to pass the dataframe to `jsonlite::write_json()` directly, and I would suggest passing the option `pretty = TRUE` to write out a human-readable JSON file. As I understand it, the use of `toJSON` with `write_json` is not correct, see https://stackoverflow.com/questions/69617597/importing-editing-and-saving-json-in-a-simple-way.

R/utils.R
```diff
@@ -36,3 +36,85 @@ calc_disc_dist_quantile <- function(prob, days, quantile) {
...
+  # convert schema to json
+  schema_json <- jsonlite::toJSON(schema, auto_unbox = TRUE)
+  jsonlite::validate(schema_json)
+
+  # save json schema
+  jsonlite::write_json(schema_json, path = write_path)
+
+  # invisibly return json schema
+  invisible(schema_json)
+}
```
pratikunterwegs, line 120: `make_json_schema` and the saving of the data dictionary as a YAML file, which is then converted to JSON only for the purposes of validating the database, seems a bit convoluted. Is this a common workflow? I would suggest saving the JSON schema as a human-readable file as well, this makes it much easier to figure out changes when reviewing. That requires passing `pretty = TRUE` as well as `auto_unbox = TRUE` to get the `$schema` string.

pratikunterwegs (commented):
Splitting my notes on some of the functions into smaller chunks.

README.Rmd
```diff
@@ -13,7 +13,7 @@ knitr::opts_chunk$set(
 )
 ```
 
-# epiparameter
+# epiparameter <img src="man/figures/epiparameter_logo.png" align="right" height="200" />
 
 <!-- badges: start -->
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
```
adamkucharski, line 19: I know updated vignettes etc. is next on list, but would be really useful to have a quick start in README with the updated syntax for pulling a parameter (e.g. pmf) out of library and defining as useable object.

R/epidist_db.R
adamkucharski: It would also be useful to have some examples of interactions with other classes in the documentation for these kinds of functions, e.g. is it possible to show a comparison of pulling out an `epidist` object from one of the database entries vs using `epidist_db`?

R/epiparam.R
adamkucharski: Following on from 1, could be helpful to have example of how to interact simply with database in the documentation, e.g. 
```
db <- epiparam()
db[db$disease == "smallpox"]
```
(or dplyr equivalent)

R/epidist.R
adamkucharski: In description, could also clarify why user might want to define an epidist object in a pipeline (rather than create database and read in). Perhaps if they've estimated parameters, then want to convert into object to pass into existing analysis steps (e.g. episoap-like templates)?

R/extract_param.R
```diff
@@ -37,7 +40,7 @@
 #' # extract parameters of a gamma distribution from median and range
 #' extract_param(
 #'   type = "range",
-#'   values = c(8, 6, 13),
+#'   values = c(10, 3, 18),
 #'   distribution = "gamma",
 #'   samples = 20
 #' )
```
adamkucharski, line 46: How is the extraction step now interacting with the database? I remember you previously mentioned it would sit somewhere between the raw database and the `epiparam` object (rather than having extracted parameters in database as before)? Or have I misunderstood?

adamkucharski (approved):

.github/workflows/test-coverage.yaml
```diff
@@ -27,5 +27,5 @@ jobs:
           needs: coverage
 
       - name: Test coverage
-        run: covr::codecov(quiet = FALSE)
+        run: covr::codecov(function_exclusions = "^make", quiet = FALSE)
```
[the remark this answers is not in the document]
joshwlambert: As both `make_json_*()` functions write to file I didn't want them running everytime the tests run and the code coverage was calculated. I could write tests and skip them if you'd prefer?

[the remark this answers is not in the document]
pratikunterwegs: One solution could be to create a temporary path from the test file, and read/write to/from that. See e.g. https://github.com/epiverse-trace/scenarios/blob/main/tests/testthat/test-import_export.R. That would allow you test all functions.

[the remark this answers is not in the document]
joshwlambert: Thanks, I'll add some tests using the temporary path.

jamesmbaazam:
I've spent some time going through the feedback already provided and think it's enough for this round of revisions.

.github/workflows/render_readme.yml
```diff
@@ -27,21 +27,21 @@ jobs:
...
           rmarkdown::render("README.Rmd")
         shell: Rscript {0}
-        
+
       - name: Commit files
         run: |
           git config --local user.email "action@github.com"
           git config --local user.name "GitHub Action"
-          git add README.md
+          git add README.md man/figures/
```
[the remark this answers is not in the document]
joshwlambert: This is to deal with adding figures that get rendered in the README. See issue #65

.github/workflows/test-coverage.yaml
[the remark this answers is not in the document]
joshwlambert: Resolved in ffff318

.github/workflows/validate-json.yaml
joshwlambert: I've updated the JSON files and functions that make them to be readable (`pretty = TRUE`), resolved in ffff318. 

For now I've removed the enum categories that were causing the issue by not properly rendering `NULL` from the yaml to the JSON. Perhaps in the future we can look into this.

R/epidist_utils.R
joshwlambert: Naming of CI elements have been changed in line with suggestions, resolved in ad66720. New examples have been added to `create_epidist_uncertainty()`, including for multiple parameters and examples for the NA and empty function call cases when uncertainty is unknown.

R/epidist_utils.R
joshwlambert: Updated documentation (title, details and return) for `create_epidist_metadata()` in e70ccd5. Added `create_epidist_region()`, which for now is not used, but is a nice idea and can be used in future to standardise the geographical setting of each study.

R/epidist_utils.R
joshwlambert: Updated documentation for `create_epidist_method_assessment()` in a1ba591. Fixed 'censor' typo throughout package.

R/epidist_utils.R
joshwlambert: Updated documentation (title and examples) for `create_epidist_summary_stats()` in a1ba591. Naming of ci fields was updated in a previous commit ad66720

R/epidist.R
joshwlambert: Point 1. ii. is an interesting one and something to be considered going forward. I do not think it is within the scope of this PR and instead have made an issue #88 to work on this further.

joshwlambert: Point 4 is something I've thought about while developing this code and something that needs addressing. I will not make changes to fix this within this PR, but have opened an issue #89 and will resolve in the near future.

pratikunterwegs: Thanks for looking into these issues - I don't have concrete suggestions so best to make new issues for further input (from users?), as you've done.

.github/workflows/validate-json.yaml
pratikunterwegs: Thanks for taking a look - I'll also give the yaml -> JSON conversion a look to see if the enum problem can be resolved somehow.

R/utils.R
pratikunterwegs: Seems resolved in https://github.com/epiverse-trace/epiparameter/commit/ffff318e5c0f595a2e4a3b6666832457f931a866

R/utils.R
pratikunterwegs: Seems resolved in https://github.com/epiverse-trace/epiparameter/commit/ffff318e5c0f595a2e4a3b6666832457f931a866

R/epidist.R
joshwlambert: Added pathogen and study to `epidist` print method. Changed distribution shorthand for name. Updated `epidist` plotting method. Updated `epidist` documentation to better explain why a user might want to create an `epidist` object using the constructor instead of the database. All resolved in 39733c7b8eed3909ae807a835b29a16b94c77b3d

R/epidist_db.R
joshwlambert: On point 4, the onset to death distribution for ebola has not been added since the update to the library in `{epiparameter}`. It will be added in the near future.

pratikunterwegs: Fair enough, will tag @thimotei here to warn that this affects {datadelay} examples and tests as well.

joshwlambert: On the first point raised, I initially didn't write this function for those reasons, however, I found the convenience of the original `epidist()` function (which read directly from the library) nice and in the scenario that someone is under a time-pressure and cannot look through the library, `epidist_db()` provides a streamlined function for accessing the library, so I'm going to keep it in the package for a while. If the users provide feedback that it is useless or impedes there ability to access data I'll consider removing.

R/epidist.R
pratikunterwegs: Get this when running `as_epidist(epiparam()[3, ])`, fitting required for params:

```r
Disease: dengue
Pathogen: dengue virus
Epi Distribution: incubation period
Study: Chan Johansson (2012) <10.1371/journal.pone.0050972>
Distribution: lognormal
Parameters:
  mu: -13.8763957698255
  sigma: 1e-10
```

and for the 11th row, no fitting required;

```r
Disease: influenza
Pathogen: influenza-A-H1N1
Epi Distribution: generation time
Study: Lessler et al. (2009) <10.1056/NEJMoa0906089>
Distribution: weibull
Parameters:
  shape: 2.36
  scale: 3.18
```

Minor quibbles about (1) title and sentence case usage, (2) missing "and" between author names, and (3) printed precision of parameters when estimated from fitting step, but best to push to next PR - may be good to make a new issue?

R/epidist_db.R
pratikunterwegs: Sounds like a good plan - was not suggesting immediate removal.

R/epidist.R
joshwlambert: Made issue (#92) to make changes in next PR.

R/epiparam_utils.R
joshwlambert: Will not address in this PR, but once issue #89 is resolved it should be resolved across the entire package. Please leave a comment on this issue if you have input.

R/epidist_db.R
joshwlambert: Updated `epidist_db()` to have a nicer error message and added an example of how different functions/classes can be used to get to the same outcome. Resolved in d5b24f13e6fcfcebc9c61f6c726f752f8696a806

R/epiparam_utils.R
pratikunterwegs: Totally fair, resolving this for now. Had not seen #89 before.

R/epidist.R
joshwlambert: Resolved in 8eb31ecde65b28f7e52e89004e3088aa3a90bf13

R/vb_epidist.R
joshwlambert: I'm not going to change the naming of the `vb_epidist` slots at the moment, but have opened an issue (#93) in order to reach a consensus on which naming is better. I'm happy to change just want to make sure it is an improvement beforehand.

joshwlambert: Resolved in 300615f84572cc791c2c006cdee61fc2a959f475.

pratikunterwegs: Thanks both fair. I've no inherent problem with the naming.

R/epiparam.R
joshwlambert: Resolved in 71187f72ab3fbf6a9e77e8841b6892e871990627

joshwlambert: Thanks for the example workflow. I will provide something similar in a vignette. I will also include examples of handling the {epiparameter} classes in a vignette using both base and tidyverse examples (noted here #94)

R/utils.R
joshwlambert: An example of the workflow of going from yaml to JSON for validation can be found here: https://github.com/covid19-forecast-hub-europe/HubValidations

R/extract_param.R
joshwlambert: The database contains only values reported in the primary literature. If summary statistics are provided but distribution parameter estimates are not, these are not added at the database compilation stage. The database should be an unaltered compilation of the literature, which is different from the previous version of the database. In the new version, if parameter estimates are not provided but adequate summary statistics are then `calc_dist_param()` will work out the best way to estimate those estimates and append them to the `epidist` object. This calculation happens automatically (by default but can be turned off) when making an `epidist` object from an `epiparam` object. The `epiparam` object when read in from the package is unchanged and so does not include parameter estimates, this could be easily changes if required.

R/epidist_db.R
```diff
@@ -0,0 +1,93 @@
...
+  disease <- match.arg(
+    arg = disease,
+    choices = unique(eparam$disease),
+    several.ok = FALSE
+  )
+
+  # filter based on pathogen and delay distribution
+  eparam <- eparam[eparam$disease == disease, ]
+
+  # extract study or default to largest sample size
```
sbfnk, line 69: Might be nice to report back to the user what has been selected and remind them to cite the study - see [here] for how this is implemented in `socialmixr`: 
https://github.com/epiforecasts/socialmixr/blob/6e508a64d830f854298867234bb9ad795f5641c8/R/load_survey.r#L149
https://github.com/epiforecasts/socialmixr/blob/main/R/cite.r

R/epidist_utils.R
```diff
@@ -0,0 +1,703 @@
...
+#' # example with uncertainty for a single parameter
+#' create_epidist_uncertainty(
+#'   ci_limits = c(1, 3),
+#'   ci = 95,
+#'   ci_type = "confidence interval"
+#' )
+#'
+#' # example for multiple parameters
+#' # lengh of list should match number of parameters
+#' list(
```
sbfnk, line 26: This way of specifying uncertainty implies that the uncertainty in parameters is uncorrelated, which it normally won't be - if that's how uncertainty is reported then obviously there is nothing that can be done about it but perhaps it would be useful to think about how it could be incorporated if it is accessible (or if the joint distribution can be estimated from raw data)?

R/epidist_utils.R
```diff
@@ -0,0 +1,703 @@
...
+#' @export
+#'
+#' @examples
+#' create_epidist_method_assessment(
+#'   censored = FALSE,
+#'   right_truncated = FALSE,
+#'   phase_bias_adjusted = FALSE
+#' )
+create_epidist_method_assessment <- function(censored = NA,
+                                             right_truncated = NA,
```
sbfnk, line 452: Could also be left truncated?

R/epiparam_utils.R
```diff
@@ -0,0 +1,447 @@
...
+        ci_limits = x$sdlog_ci_limits,
+        ci = x$sdlog_ci,
+        ci_type = ifelse(
+          test = x$inference_method == "mle",
+          yes = "confidence interval",
+          no = "credible interval"
+        )
+      )
+    )
+  } else if (x$prob_distribution %in% "negative_binomial") {
```
sbfnk, line 96: might want to consider using R abbreviations, i.e. "lnorm", "nbinom" etc.

R/calc_dist_params.R
```diff
@@ -0,0 +1,159 @@
...
+
+  if (!anyNA(mean_sd)) {
+    # make mean_sd a class with the name of the prob dist for multiple dispatch
+    prob_dist_params <- convert_params(
+      summary_stats = summary_stats,
+      prob_dist = prob_dist
+    )
+  } else if (!all(is.na(lower_quantiles)) && !all(is.na(upper_quantiles))) {
+    # use the first numeric quantile in the set to calculate the parameters
+    lower_percentile <- lower_quantiles[!is.na(lower_quantiles)][1]
```
sbfnk, line 75: Do you need to check that they're symmetric? I think otherwise you might want to consider fitting with weighted least squares as the tail percentiles have more uncertainty (and similarly for fitting the ranges). I think it might be good document somewhere how the fitting is done as I for one didn't find it easy to follow in the code (and still don't fully understand it - but that might be just me).

R/calc_dist_params.R
```diff
@@ -0,0 +1,159 @@
...
+#' @description The summary_stats object needs to have a class attribute with
+#' the name of the probability distribution in order for the correct conversion
+#' function to be dispatched.
+#'
+#' @inheritParams new_epidist
+#' @inheritParams epidist
+#'
+#' @return Named vector
+#' @keywords internal
+convert_params <- function(summary_stats,
```
sbfnk, line 133: Might be nice in the future to extend to more conversions, e.g. could also estimate lognormal parameters using the median instead of the mean.

sbfnk (commented):
Really nice work!

Bisaloo (commented):
I mostly reviewed the code as I went through it but didn't review if the entire structure makes sense yet. I'd rather do it one this is merged to `main` and you have created a pkgdown index: https://pkgdown.r-lib.org/reference/build_reference.html#reference-index

R/bind_epiparam.R
```diff
@@ -0,0 +1,68 @@
...
+    validate_epiparam(epi_obj)
+    # epiparam objects can be directly binded
+    out <- rbind(epiparam, epi_obj)
+    # validate new epiparam object
+    validate_epiparam(out)
+    return(out)
+  } else if (is.data.frame(epi_obj)) {
+    stopifnot(
+      "data frame provided must have the same column names as epiparam" =
+      identical(colnames(epi_obj), colnames(epiparam))
```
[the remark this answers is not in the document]
joshwlambert: I think the column reordering is automatically handled by `rbind()` given the column names. 

``` r
df1 <- data.frame(a = 1, b = 2, c = 3)
df2 <- data.frame(c = 1, b = 2, a = 3)
df3 <- data.frame(b = 1, a = 3, c = 2)
setequal(colnames(df1), colnames(df2))
#> [1] TRUE
setequal(colnames(df1), colnames(df3))
#> [1] TRUE
rbind(df1, df2, df3)
#>   a b c
#> 1 1 2 3
#> 2 3 2 1
#> 3 3 1 2
```

<sup>Created on 2023-02-13 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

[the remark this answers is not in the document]
joshwlambert: Resolved in bc744a03f350d4a63ffdc4d2540df6ae48580c92

R/bind_epiparam.R
```diff
@@ -0,0 +1,68 @@
...
+  } else if (is_vb_epidist(epi_obj)) {
+    validate_vb_epidist(epi_obj)
+  } else if (is_epiparam(epi_obj)) {
+    validate_epiparam(epi_obj)
+    # epiparam objects can be directly binded
+    out <- rbind(epiparam, epi_obj)
+    # validate new epiparam object
+    validate_epiparam(out)
+    return(out)
+  } else if (is.data.frame(epi_obj)) {
```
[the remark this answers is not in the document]
joshwlambert: I've taken a look at splitting this up into individual methods but it increases the amount of code and code repetition, therefore, for now I will keep the current implementation.

R/check_epidist.R
```diff
@@ -0,0 +1,67 @@
+#' Checks the probability distribution name and parameters to be used in an
+#' `epidist` object
+#'
+#' @inheritParams new_epidist
+#'
+#' @return Nothing, errors if checks fail
```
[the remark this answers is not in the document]
joshwlambert: Resolved in e895e158809fc6550c65ff9ff71086638b77ab37

R/check_epidist.R
```diff
@@ -0,0 +1,67 @@
...
+  # return probability distribution parameters
+  invisible(prob_dist_params)
+}
+
+#' Checks the parameter uncertainty list in an `epidist` object
+#'
+#' @inheritParams new_epidist
+#' @inheritParams epidist
+#'
+#' @return Nothing, errors if checks fail
```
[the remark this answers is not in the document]
joshwlambert: Resolved in e895e158809fc6550c65ff9ff71086638b77ab37

R/vb_epidist.R
```diff
@@ -0,0 +1,344 @@
...
+print.vb_epidist <- function(x, ...) {
+  format(x, ...)
+}
+
+#' Format method for vb_epidist class
+#'
+#' @param x vb_epidist object
+#' @param ... further arguments passed to or from other methods
+#'
+#' @return Nothing (prints output)
```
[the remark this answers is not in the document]
joshwlambert: Resolved in e895e158809fc6550c65ff9ff71086638b77ab37

R/create_prob_dist.R
```diff
@@ -0,0 +1,124 @@
...
+        lambda = unname(prob_dist_params)
+      )
+    )
+  }
+
+  # apply truncation to distribution if specified
+  if (!is.na(truncation)) {
+    if (discretise) {
+      warning(
+        "Truncation is not yet implemented for discrete distributions",
```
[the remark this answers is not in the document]
joshwlambert: Resolved in 9c89a44c5bebf1906dceba1f61197b54aea76448

R/epidist_utils.R
```diff
@@ -0,0 +1,703 @@
...
+#' of the distribution for the `epidist` object
+#'
+#' @param ci_limits A numeric vector of length two with the lower and upper bound of
+#' the confidence interval or credible interval
+#' @param ci A numeric specifying the interval for the ci, e.g. 95 is
+#' 95% ci
+#' @param ci_type A character string, either "confidence interval" or "credible
+#' interval"
+#'
+#' @return List of three elements
```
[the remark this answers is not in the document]
joshwlambert: Resolved in 6531c0e03d7404d06c023506eea387e22064476a

R/epidist_utils.R
```diff
@@ -0,0 +1,703 @@
...
+#' create_epidist_citation(
+#'   author = "Smith_etal",
+#'   year = 2002,
+#'   DOI = "10.19832/j.1366-9516.2012.09147.x"
+#' )
+create_epidist_citation <- function(author = NA_character_,
+                                    year = NA_integer_,
+                                    PMID = NA_character_,
+                                    DOI = NA_character_,
+                                    use_PMID = FALSE) {
```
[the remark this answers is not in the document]
joshwlambert: I had written it this way to not clutter the reference with the PMID, but happy to always include it when available.

R/epidist_db.R
joshwlambert: Thanks for pointing this out. I will implement this in another pull request and have logged it here #97.

NAMESPACE
```diff
@@ -1,17 +1,66 @@
...
+export(epiparam)
 export(extract_param)
 export(extract_param_percentile)
 export(extract_param_range)
 export(gamma_meansd2shapescale)
 export(gamma_shapescale2meansd)
+export(generate)
+export(geometric_mean2prob)
+export(geometric_prob2mean)
+export(head)
```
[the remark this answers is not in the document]
joshwlambert: This should now be resolved with the correct importing of S3 generics.

R/epidist_utils.R
```diff
@@ -0,0 +1,703 @@
...
+                                         dispersion_ci = NA_real_,
+                                         lower_range = NA_real_,
+                                         upper_range = NA_real_,
+                                         q_025 = NA_real_,
+                                         q_05 = NA_real_,
+                                         q_25 = NA_real_,
+                                         q_75 = NA_real_,
+                                         q_875 = NA_real_,
+                                         q_95 = NA_real_,
+                                         q_975 = NA_real_) {
```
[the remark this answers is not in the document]
joshwlambert: These quantiles were chosen as they are the most commonly reported in the literature. For now I will keep the database the same with these quantiles, but given Seb's suggestion in this PR I will change it so that quantiles are provided by the user instead of being predefined.

R/calc_dist_params.R
joshwlambert: I will continue to add conversions functions as we develop. For now I've opened an issue (#98), and suggested a new way to format the conversion functions. Let me know what you think.

R/epidist_utils.R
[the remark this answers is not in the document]
joshwlambert: Resolved in b619cd21359b1f51610b910b842b14c4c6c00182

R/epiparam.R
```diff
@@ -0,0 +1,304 @@
...
+       num_delay_dist = num_delay_dist,
+       num_offspring_dist = num_offspring_dist,
+       num_studies = num_studies,
+       num_continuous_distributions = num_cont_dist,
+       num_discrete_distributions = num_disc_dist,
+       num_vector_borne_diseases = num_vector_borne
+  )
+}
+
+#' head method for epiparam class
```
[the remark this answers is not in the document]
joshwlambert: When I removed `head()` and `tail()` the print methods failed on [L208](https://github.com/epiverse-trace/epiparameter/blob/epiparams/R/epiparam.R#L208), because of a subsetting issue. I'm not sure why this is happening, but I remember before adding the `head()` and `tail()` methods it was not dispatching the `data.frame` method as expected, which was the original reason I added these functions. If there is a better way to set this up (perhaps using `NextMethod()` somewhere) in order to remove `head()` and `tail()` please let me know.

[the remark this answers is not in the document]
pratikunterwegs: Chipping in since I adapted this for the revised implementation of `scenario`, you could copy the implementation of `head()` to L208, so that you convert from `epiparam` to `data.frame`, on which you should be able to safely call `head()` with subsetting.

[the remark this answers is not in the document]
joshwlambert: But there would still be a `head()` `epiparam` method implemented and it would break the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) rule, so I don't see the difference. Am I missing something?

R/epiparam_utils.R
```diff
@@ -0,0 +1,447 @@
...
+#' as_epiparam(edist)
+as_epiparam <- function(x) {
+
+  # for vb_epidist or list of epidists call as_epiparam recursively
+  if (!is_epidist(x)) {
+    eparam <- data.frame()
+    for (i in seq_along(x)) {
+      eparam_row <- as_epiparam(x[[i]])
+      eparam <- rbind(eparam, eparam_row)
+    }
```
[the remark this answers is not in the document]
joshwlambert: Resolved in 9413c2e4834add054cf82233d0d0a33fddb49946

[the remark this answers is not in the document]
joshwlambert: The speed difference between the old implementation: 
```
Unit: milliseconds
                                    expr      min       lq     mean   median       uq      max neval
 res <- as_epiparam(edist, fast = FALSE) 324.9551 333.2591 336.9286 334.0796 335.5553 434.5644   100
```

and the new implementation:
```
Unit: milliseconds
                                   expr     min       lq     mean   median       uq      max neval
 res <- as_epiparam(edist, fast = TRUE) 280.423 283.1359 289.2464 285.7036 290.2664 453.6495   100
```

is relatively small on lists of length 100. 

After reading https://stackoverflow.com/questions/14693956/how-can-i-prevent-rbind-from-geting-really-slow-as-dataframe-grows-larger I also tried the `do.call(rbind, ...)` option but it was not quicker than the new implementation.

R/epiparam.R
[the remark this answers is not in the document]
pratikunterwegs: No, you can remove the `head()` implementation if you update L208 to be `print(utils::head(as.data.frame(x))[c("disease", "epi_distribution", "prob_distribution")])`.

R/epiparam_utils.R
joshwlambert: This will be changed in an upcoming pull request. The discussion of naming conventions is here #86

R/epidist_utils.R
joshwlambert: From the papers used to collated parameters for the database I have not seen any report left truncation. However, I guess it is feasible that the delay distributions could be left-truncated so I will think about how best to implement this change. `left_truncated` can be easily added to `create_epidist_method_assessment()` but there are knock-on effects as there is currently only one column for truncation (assumed right truncation) in the database, so adding left truncation will also cause a change to the database.

R/epiparam.R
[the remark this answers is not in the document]
joshwlambert: Yes, but before I added `head.epiparam()` the `head()` generic was not dispatching `head.data.frame()`, and instead I was getting the first 6 rows the database printed using `print.epiparam()`. Instead I want the `head.epiparam()` to print as though it were a data frame.

I'm happy to take a look at `{scenarios}` and discuss another way to implement.

R/epidist_utils.R
sbfnk: Serial intervals are often left truncated at 0 (implicitly excluding cases where A infects B but B develops symptoms before A). Here's an example where they were left truncated at 2 (cases with symptom onset 0 or 1 days apart were classified as co-primary): https://www.eurosurveillance.org/content/10.2807/1560-7917.ES.2022.27.15.2001551

R/epiparam.R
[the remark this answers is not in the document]
Bisaloo: If I understand correctly your problem, everything is working as designed.

The `head()` method from `matrix`, `data.frame` and `array` will return an object of the same class as `x`.

So if you're calling `head.data.frame()` on an `epiparam` object, you get an `epiparam` object back, and it's then dispatched to `print.epiparam()` when you print it.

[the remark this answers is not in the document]
Bisaloo: Unclassing your object while calling `head()` or `tail()` seems like a potential source of errors or confusion for your users.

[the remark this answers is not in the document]
joshwlambert: I agree. I wanted `head()` and `tail()` to print all columns which is not the case for `print.epiparam()`. If you think it is bad practise to use `as.data.frame()` in the `head()` and `tail()` methods I could rename the functions to make clear that they will return a data frame and not use the generic.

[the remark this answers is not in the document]
pratikunterwegs: I think the issue is that the `print()` method relies on the `format()` method, which calls `print(head(...))`. So `head()` returns an `epiparam` (or in my case, `scenario`), and then looks for a `print()` method, which in turn calls `head()` and so on. Calling `NextMethod(generic = "print", object = x)` does't really seem to work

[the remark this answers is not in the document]
Bisaloo: > I agree. I wanted head() and tail() to print all columns which is not the case for print.epiparam(). If you think it is bad practise to use as.data.frame() in the head() and tail() methods I could rename the functions to make clear that they will return a data frame and not use the generic.

I'm not sure. The problem sounds more like you're not happy with the way your `print()` works. If you like the output is some situations but not in others, you may not be looking for `print()`. If you don't like it for the output of `head()`, chances are you won't like it in some other legitimate situations. Maybe what you put in `print.epiparam()` should really be `summary.epiparam()` and you should stick with `print.data.frame()` for printing.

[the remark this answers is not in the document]
joshwlambert: I like the current `print.epiparam()` implementation (which uses `summary.epiparam()`) as I did not want all 56 columns of the `epiparam` object printed. One option would be to use `tibble` to provide a cleaner print method but I didn't want the dependency. For now I'll keep with the print method as it's minimal and informative. If you both think it's best to have the `head()` and `tail()` print as `epiparam` by default then I'm happy to change and I will write two new functions that do what the current `head()` and `tail()` are doing.

R/epiparam.R
```diff
@@ -0,0 +1,304 @@
...
+      if (all(is.na(x))) c(NA_real_, NA_real_) else as.numeric(x)
+    })
+    params[[i]] <- split_ci
+  }
+
+  # convert NAs to correct type
+  na_col_index <- which(
+    vapply(params, function(x) all(is.na(x)), FUN.VALUE = logical(1))
+  )
+  params[, na_col_index] <- NA_real_
```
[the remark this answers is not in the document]
joshwlambert: I can't remember exactly why I added this. I thought there may have been some issues with type conversion (so I made the `NA`s numeric), however, playing around with it now I don't find any different behaviour if the `NA`s are logical or numeric. So I've removed this code chunk in a4fbb5130ba8d2f82614fafd12df954e348844fe

R/calc_dist_params.R
```diff
@@ -0,0 +1,159 @@
...
+          x = names(lower_percentile),
+          fixed = TRUE
+        ),
+        gsub(
+          pattern = "q_",
+          replacement = "",
+          x = names(upper_percentile),
+          fixed = TRUE
+        )
+    ))
```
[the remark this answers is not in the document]
joshwlambert: Thanks for spotting this. There was a bug which I think I've fixed in 5b1d12c47080d813f9eba6bfc36447dbeda3e5f1. I'm not really happy with how the percentile names are being handled, which is currently using regex. Will take a look into this and try and improve soon. I think the naming is partly at fault, having decimal points in the names would help.

R/epidist_utils.R
joshwlambert: I had overlooked this aspect. I think most papers report uncertainty without specifying a correlation between them. What structure would be best for this, a n-dim matrix with n parameters?

R/epidist.R
```diff
@@ -1,242 +1,693 @@
...
+    prob_dist = prob_dist,
+    prob_dist_params = prob_dist_params,
+    discretise = discretise,
+    truncation = truncation
+  )
 
-  if (nrow(params) == 0) {
-    stop("Need to select pathogen and distribution in the dataset")
+  # calculate distribution quantiles and add to epidist
+  quants <- c(0.025, 0.05, 0.25, 0.5, 0.75, 0.875, 0.95, 0.975)
```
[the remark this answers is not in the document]
joshwlambert: Updated the quantiles in 46ff88e78c2ca13ae2a4b29c43b60e8a6c35d72c to make them flexible to different user input.

R/epidist.R
```diff
@@ -1,242 +1,693 @@
...
+  if (isTRUE(discretise)) {
+    quantiles <- prob_dist$q(quants)
+  } else {
+    quantiles <- stats::quantile(prob_dist, quants)[[1]]
   }
+  names(quantiles) <- c(
```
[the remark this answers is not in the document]
joshwlambert: Resolved in 46ff88e78c2ca13ae2a4b29c43b60e8a6c35d72c

R/epidist_utils.R
```diff
@@ -0,0 +1,703 @@
...
+#'     ci_type = "confidence interval"
+#'   )
+#' )
+#'
+#' # example with unknown uncertainty
+#' # the function can be called without arguments
+#' create_epidist_uncertainty()
+#' # or give NA as the first argument
+#' create_epidist_uncertainty(NA)
+create_epidist_uncertainty <- function(ci_limits = NA_real_, ci, ci_type) {
```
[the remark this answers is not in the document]
joshwlambert: I've made an issue (#99) for this topic. I think dedicating a pull request to better handling of uncertainty is required, also to prevent more commits in the current PR.

pratikunterwegs (approved):
Thanks @joshwlambert, looks good to me overall, and happy to approve. There are a few open conversations though, which seem to be optional suggestions, and I would recommend converting them to new issues to be addressed in a different PR.

Putting down a few things I noticed which might be good to convert to issues or keep in mind for future fixes:

1. The vignette chunks are set to `eval = FALSE` since the functions are out of date with the current PR - something to keep in mind.
2. There are over 1,300 tests which is potentially unwieldy - suggest eventually compressing tests - for example, a single key-value pair comparison (via a snapshot) for the column types in the parameter database, rather than a single `expect` for each column.
3. L.209 in `epiparam_utils.R` has a hardcoded number of columns.
4. Might be good to revisit the function `get_percentiles()` - haven't tested it but getting names and converting to decimal values, as well as determining the upper and lower percentile values seems a little hacky at the moment - am I right in thinking one value has to be > 50 and one < 50? An understandable choice, but not mentioned in the documentation. These lines in the tests suggests this is the case: https://github.com/epiverse-trace/epiparameter/blob/13c4f40f597700880dc16875599baf19bd4dc752/tests/testthat/test-get_percentiles.R#L8-L9.

README.Rmd
joshwlambert: Resolved in 608ffecf13523ae0e18cc333e98745e1597404f0

R/epidist_utils.R
joshwlambert: Issue opened for further work (#101).

R/calc_dist_params.R
joshwlambert: This is a good point. I will discuss with @adamkucharski and improve the documentation of parameter extraction and how it is used to automatically calculate parameters in `calc_dist_params()` (#102).

PR #96, opened 2023-02-16
Delete lint-changed-files.yaml

Bisaloo:
since it is now handled at the org level


joshwlambert (approved):

joshwlambert:
@Bisaloo thanks for handling the worflows!

Issue #100, opened 2023-03-01
Make it possible to pipe `epiparam()` objects


jamesmbaazam:
**Is your feature request related to a problem? Please describe.**

I am imagining that `epiparameter` would typically be used in piped operations but I am getting an error because `epiparam()` objects are not pipeable. 

**Describe the solution you'd like**
Make `epiparam()` objects pipeable.

**Additional context**

Here is an example where I was trying to find the unique diseases in the database, but got an error.

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
a <- epiparameter::epiparam()
a %>% dplyr::distinct(disease)
#> Error in validate_epiparam(NextMethod()): epiparam object does not contain the correct columns
```

<sup>Created on 2023-03-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

avallecam:
Since the epiparam object is of class `data.frame` in the background, we can use `dplyr::as_tibble()` before using other `dplyr` functions:

``` r
library(epiparameter)
library(tidyverse)
a <- epiparameter::epiparam()
class(a)
#> [1] "epiparam"   "data.frame"
a %>% 
  as_tibble() %>% 
  dplyr::distinct(disease)
#> # A tibble: 24 × 1
#>    disease                      
#>    <chr>                        
#>  1 Adenovirus                   
#>  2 Chikungunya                  
#>  3 COVID-19                     
#>  4 Dengue                       
#>  5 Ebola Virus Disease          
#>  6 Hantavirus Pulmonary Syndrome
#>  7 Human Coronavirus            
#>  8 Influenza                    
#>  9 Japanese Encephalitis        
#> 10 Marburg Virus Disease        
#> # … with 14 more rows
```

<sup>Created on 2023-03-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

avallecam:
Although, in #111 I used `dplyr::filter()` directly after the `epiparam()` output and it worked nicely.

avallecam:
Also related, I just encountered the [Editorial decisions of the Epi R Handbook](https://www.epirhandbook.com/en/editorial-and-technical-notes.html#editorial-decisions).

<html><body>


Subject | Considered | Outcome | Brief rationale
-- | -- | -- | --
[...]

PR #125, opened 2023-03-23
Convert `epiparam` to `data.frame` instead of erroring

joshwlambert:
This PR addresses issue #100, in which it was described how the previous implementation of `<epiparam>` subsetting was preventing easy exploration of the {epiparameter} library/database. Changes in this PR alter this behaviour to return a `<data.frame>` when the class is invalidated. 

This involved:
- adding {dplyr} and {rlang} to Suggests in the DESCRIPTION
- adding a `reconstruct` argument to `validate_epiparam()` to allow class independent validation (i.e. data frames can be checked as valid `epiparam` objects)
- `names()` and `$` setters for `epiparam` were added
- Reconstruction functions were added
    - `epiparam_reconstruct()`
    - `epiparam_can_reconstruct()`
    - `df_reconstruct()`
    - `dplyr_reconstruct.epiparam()`
- `s3_register()` was taken from {rlang} (code is unlicensed) to register `dplyr::dplyr_reconstruct()` without importing the package
- `.onLoad()` was added to register method using `s3_register()`
- Updates to documentation and tests due to new functionality


joshwlambert:
Assigning @pratikunterwegs to review this PR as well, due to possible overlap with developments in [{scenarios}](https://github.com/epiverse-trace/scenarios).

Bisaloo:
Hi @joshwlambert, thanks for this great change! I believe this will be a significant improvement.

Rather than using the `.onLoad()` workaround that is / will be deprecated very soon, how would you feel about already bumping the R version requirement to 3.6.0? We would be making this change only 3 weeks ahead of the tidyverse so I think it's quite safe.

jamesmbaazam:
Thanks for tagging me, Josh. I will take a look asap.

R/epiparam_utils.R
```diff
@@ -438,14 +438,113 @@ as_epiparam <- function(x) {
...
+#' @param to The reference object, in this case an `epiparam` object
+#'
+#' @return An `epiparam` object (if the input is valid) or a `data.frame`
+#' @keywords internal
+epiparam_reconstruct <- function(x, to) {
+  if (epiparam_can_reconstruct(x)) {
+    df_reconstruct(x, to)
+  } else {
+    x <- as.data.frame(x)
+    message("Removing crucial column in `<epiparam>` returning `<data.frame>`")
```
pratikunterwegs, line 505: This could potentially be bumped up to a warning.

pratikunterwegs (approved):
Thanks for tagging @joshwlambert, this looks good, and seems to be very thoroughly thought through from what I can see. I can see {scenarios} adopting most of these features as well.

I think this should be fine to merge as it is. I tried out the new functionality, including how it interacts with functions such as `bind_epiparam()`, `as_epidist()`, and `dplyr::filter()`. Seems to all work as intended. I have only one suggestion on perhaps bumping the message when an `epiparam` is converted to a `data.frame` up to a warning, but I leave that to you.

R/epiparam_utils.R
joshwlambert: In tidyverse and tidymodels packages (e.g. [^1]) this conversion is actually silent. I don't think it warrants a warning because it may be the users intention to subset out a certain column which would trigger this class downgrading. I'll leave it for now and if consensus from users builds that this should be a warning then I'll change in the future.

[^1]: https://github.com/tidymodels/dials/blob/8de943905ef0c5bcedcc7154a86c239623957d46/R/compat-vctrs-helpers-parameters.R

joshwlambert:
@pratikunterwegs Thanks a lot for reviewing. Glad to hear the info was helpful for potential {scenarios} development as well.

R/epiparam_utils.R
pratikunterwegs: That's fair, and I leave it to you - I don't have a good sense of whether users would be confused by class downgrading or take it in their stride, such as when they pipe a downgraded `epiparam` to an `epiparam` method and can't identify what causes the resulting error - as you say probably good to leave it for now and if this comes up frequently as an issue, to take another look.

joshwlambert:
@jamesmbaazam @avallecam @Bisaloo I will merge this PR today. If there is anything relating to this PR that you'd like to raise please feel free to open an issue.

jamesmbaazam:
Sorry I couldn't get to this PR in time @joshwlambert. I will test it and raise issues as suggested.

jamesmbaazam:
One more comment: I think this would be a good opportunity to [tag](https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/managing-commits/managing-tags) this change (or patch release it??) to make it easier to refer to and test. 

What do you think?

Issue #127, opened 2023-03-24
Add `parameters()` method to `<epidist>` class


joshwlambert:
Once an `<epidist>` object is constructed, accessing each of its elements can be easily done with `$` (it is just a list with a class attribute fundamentally). However, the `$prob_dist` element is itself an S3 class, either from {distributional} or {distcrete}. Is it not obvious how to access the parameters from these objects, and in the case of {distributional} objects it requires a call to a function. 

```r
unlist(distributional::parameters(onset_to_death_ebola$prob_dist))
```

Adding a `parameters()` method to `<epidist>` will ensure it is easy to extract the distribution parameters from the object in a consistent way independent of whether the probability distribution is discretised or not.

This can inherit the S3 generic `parameters()` from {distributional}.

Issue #124, opened 2023-03-22
Discretise `<epidist>`with continuous distribution 


joshwlambert:
If an `<epidist>` with a continuous distribution is loaded, but the PMF is required, there is currently no method to discretise the distribution. See example of continuous `<epidist>` below.

```r
onset_to_death_ebola <- epiparameter::epidist_db(
  disease = "Ebola Virus Disease", 
  epi_dist = "onset_to_death",
  author = "Barry_etal")
```

I propose adding an `<epidist>` class method that converts the `$prob_dist` element of the `<epidist>` from using {distributional} to {distcrete}, while preserving the parameter information.

PR #128, opened 2023-03-30
Added `discretise()` and `parameters()` methods to `<epidist>`

joshwlambert:
This pull request addresses issues #124 & #127. Two new function are added:
- `discretise()` 
- `parameters()`

`discretise()` takes an `<epidist>` object with a continuous distribution (e.g. gamma or lognormal) and converts it to its discretised form. This can be useful when an entry in the database is continuous but a discretised distribution is required for modelling. This function requires adding a new generic.

`parameters()` is a method that extracts the parameters of the probability distribution and returns them in a consistent manner (named vector). This is useful as the functions and accessors differ depending on whether the distribution is discretised (using {distcrete}) or not (using {distributional}), and now the user does not have to remember these differences when working with `<epidist>` objects. This function does not require adding a new generic as it imports the generic from the {distributional} package.

Tests and documentation are updated in accordance with the above-mentioned changes. Other {epiparameter} functions have been updated to use the newly added functions where necessary.


R/epidist.R
```diff
@@ -429,7 +429,7 @@ format.epidist <- function(x, header = TRUE, vb = NULL, ...) {
   } else {
 
     # decide on parameter format from magnitude of number
-    params <- unlist(distributional::parameters(x$prob_dist))
+    params <- parameters(x)
     format_params <- ifelse(
       test = any(params > 9.999e-3 & params < 1e4),
```
adamkucharski, line 434: Presumably this won't break anywhere (e.g. if superspreading has very large MLE for k)?

R/epidist.R
```diff
@@ -704,3 +704,129 @@ generate.epidist <- function(x, times, ...) {
...
+#'   prob_distribution_params = c(shape = 1, scale = 1)
+#' )
+#' discretise(ebola_incubation)
+discretise <- function(x, ...) {
+  UseMethod("discretise")
+}
+
+#' @rdname discretise
+#' @export
+discretise.epidist <- function(x, ...) {
```
adamkucharski, line 734: Worth flagging in function description (or somewhere else) that distributional functions may no longer apply to distcrete objects (e.g. `generate(onset_to_death_ebola$prob_dist,10)` vs
`generate(epiparameter::discretise(onset_to_death_ebola)$prob_dist,10)`? Not sure what relevant syntax is for same random draw operation with `distcrete` (and corresponding pdf/pmf functions).

adamkucharski (commented):
Have added a couple of comments about potential for things to break in downstream use. If not already planned, would also be useful to update epiparameter vignette to give some easy quick start examples for how to output random draws, density functions (rnorm, dnorm etc.) with these classes.

R/epidist.R
joshwlambert: It should work fine. Here's an example of an `<epidist>` object for superspreading (negative binomial) with large and small values for k.

``` r
library(epiparameter)
transmission <- epidist(
  disease = "SARS", 
  epi_dist = "offspring_distribution", 
  prob_distribution = "nbinom", 
  prob_distribution_params = c(mean = 1, dispersion = 1e10)
)
#> Citation cannot be created as either author, year or DOI is missing
transmission
#> Disease: SARS
#> Pathogen: NA
#> Epi Distribution: offspring distribution
#> Study: No citation available
#> Distribution: negbin
#> Parameters:
#>   n: 10000000000.000
#>   p: 1.000

transmission <- epidist(
  disease = "SARS", 
  epi_dist = "offspring_distribution", 
  prob_distribution = "nbinom", 
  prob_distribution_params = c(mean = 1, dispersion = 1e-10)
)
#> Citation cannot be created as either author, year or DOI is missing
transmission
#> Disease: SARS
#> Pathogen: NA
#> Epi Distribution: offspring distribution
#> Study: No citation available
#> Distribution: negbin
#> Parameters:
#>   n: 1e-10
#>   p: 1e-10
```

<sup>Created on 2023-04-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/epidist.R
joshwlambert: The distribution functions in {epiparameter} are setup to handle both discretised and non-discretised distributions:

``` r
library(epiparameter)
library(distributional)
ebola_si <- epidist(
  disease = "ebola", 
  epi_dist = "serial_interval", 
  prob_distribution = "gamma", 
  prob_distribution_params = c(shape = 1, scale = 1)
)
#> Citation cannot be created as either author, year or DOI is missing
ebola_si
#> Disease: ebola
#> Pathogen: NA
#> Epi Distribution: serial interval
#> Study: No citation available
#> Distribution: gamma
#> Parameters:
#>   shape: 1.000
#>   rate: 1.000

cdf(ebola_si, q = 50)
#> [1] 1
generate(ebola_si, times = 10)
#>  [1] 0.2634135 1.1430287 1.0172865 0.4948622 1.4337434 0.8281407 0.5926481
#>  [8] 0.1577783 0.1523971 0.3412915
density(ebola_si, at = 0.5)
#> [1] 0.6065307
quantile(ebola_si, p = 0.5)
#> [1] 0.6931472

ebola_si <- discretise(ebola_si)
ebola_si
#> Disease: ebola
#> Pathogen: NA
#> Epi Distribution: serial interval
#> Study: No citation available
#> Distribution: discrete gamma
#> Parameters:
#>   shape: 1.000
#>   scale: 1.000

cdf(ebola_si, q = 50)
#> [1] 1
generate(ebola_si, times = 10)
#>  [1] 2 2 2 1 1 1 2 1 2 5
density(ebola_si, at = 0.5)
#> [1] 0.3934693
quantile(ebola_si, p = 0.5)
#> [1] 0
```

<sup>Created on 2023-04-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

But I definitely agree we need to make this clear in the documentation and vignettes.

R/epidist.R
adamkucharski: Thanks. Although I notice it seems to be converting (mean,dispersion) to (n,p) in epidist object here?

adamkucharski (approved):

R/epidist.R
joshwlambert: Yes, that's due to the use of the {distributional} package underlying the {epiparameter} print method. We could write a conversion into the print method to make sure the parameters printed are the same as those input into the `<epidist>`, in this case mean and dispersion. 

What do you think?

Issue #126, opened 2023-03-24
Create JSON data and schema within validation workflow


joshwlambert:
Currently, in order to perform JSON validation using the validate-json.yaml Github action workflow, a JSON version of the csv database and a JSON version (schema) of the data dictionary are stored in the package. However, as these are only used in the validation, and there are functions to created them easily (`make_json_data()` and `make_json_schema()`) these can be made on the fly when the workflow executes and not stored in the package, having the added benefit of making the package more lightweight.

joshwlambert:
This issue was resolved in #129.

PR #129, opened 2023-03-30
JSON validation workflow generates JSON files

joshwlambert:
This pull request addresses issue #126. 

Currently, JSON validation (via Github actions) uses saved JSON data and schema files in the {epiparameter} package. However, these are only used for validation and therefore can be generated during the validation workflow. The JSON data and schema files have been deleted in this PR, making the package more lightweight.

It also removes the need for JSON conversion functions (`make_json_data()` and `make_json_schema()`). 

The tests for these functions have also been removed.


joshwlambert:
@Bisaloo I propose to merge this PR as is. Some of the commits were trying to fix the workflow, as I couldn't getting working straight away. I tagged you mainly to check if the workflow can be improved as the changes I've made are rather basic. I can see the workflow is working correctly, so don't feel the review is necessary. 

As some of the commits are patches attempting to get the workflow to run, I propose rebasing and squash merging to not populate the commit history with attempts to fix the yaml.

.github/workflows/validate-json.yaml
```diff
@@ -18,18 +18,53 @@ jobs:
         with:
           use-public-rspm: true
 
-      - name: Install V8
-        run: sudo apt install libv8-dev
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          extra-packages:
```
Bisaloo, line 23:
```suggestion
          packages:
```

Since you don't need to install epiparameter or any of its dependencies

Issue #75, opened 2022-12-02
Split vignette into topic-specific vignettes


joshwlambert:
The current vignette is relatively large and contains information on both the functionality of the `epiparameter` package and its integration with other packages. This can be confusing to people new to the package, therefore I suggest splitting the vignette into 1) core functionality (one vignette) and 2) applications (several vignettes). 

Remove old vignette:
- [x] remove vignette

Core functionality: 
- [x] Getting started (introduction to `<epidist>` and `<epiparam>` classes)
- [ ] Using `<epidist>` and `<epiparam>` methods
- [x] Parameter conversion and extraction 

Applications: 
- [ ] Quarantine 
- [ ] Transmissibility
- [ ] Case-fatality rate
- [ ] Superspreading

Auxiliary:
- [x] Parameter extraction bias

joshwlambert:
Instead of having one vignette per application, I will add a use-cases vignette which will include the topics included in the above description, but will not be extensive/comprehensive. 

For more expansive applications of {epiparameter} the [Epiverse-TRACE training materials](https://github.com/epiverse-trace/tutorials) (which are currently being worked on and will include more information on applying {epiparameter} in the future) can be used.

joshwlambert:
There is a now an ecosystem of documentation, [training](https://github.com/epiverse-trace/tutorials-middle) and [how to guides](https://github.com/epiverse-trace/howto) within Epiverse-TRACE. Therefore, the application of {epiparameter} to other packages in the ecosystem or epidemiological tasks is no longer viewed necessary to be in this package. 

Usage of {epiparameter} is included in other Epiverse-TRACE R packages, such as {superspreading} and {simulist}.

The suggested vignette on "Using `<epidist>` and `<epiparam>` methods" is now outdated as the `<epiparam>` class has been removed since #197 and the `<epidist>` class has been renamed `<epiparameter>` since #360.

Closing this issue, if vignettes outlining the application of {epiparameter} to epidemiological tasks are reconsidered to be needed this issue can be reopened.

PR #136, opened 2023-04-17
Added new introductory vignette

joshwlambert:
Added new getting started vignette to give a brief overview of the {epiparameter} package, including functions and classes.

This is part of issue #75 which aims to change from having one or a few large vignettes to having several shorter topic-specific vignettes. PR #132 removed the original introductory vignette which included information on both the {epiparameter} package and its applications. 

This new vignette is exclusively describing the functionality and data in {epiparameter}. For other upcoming vignettes, see checklist in #75.

### DIFF
--- a/vignettes/epiparameter.Rmd
+++ b/vignettes/epiparameter.Rmd
@@ -83,6 +83,9 @@ To see a full list of the diseases and distributions stored in the library use t
 head(list_distributions(epi_dist_db))
 ```
 
+More details on the data collation and the library of parameters can be found in
+the [Data Collation and Synthesis Protocol vignette](https://epiverse-trace.github.io/epiparameter/articles/data_protocol.html).
+
 ### Single set of epidemiolgical parameters
 
 The second class introduced in the {epiparameter} package is the `<epidist>` class. This holds a single set of epidemiological parameters. 
--- a/vignettes/epiparameter.Rmd
+++ b/vignettes/epiparameter.Rmd
@@ -29,6 +29,8 @@ An outbreak of a known or potentially novel pathogen is detected and the paramet
 This vignette will provide a introduction to the data stored within {epiparameter}, how to read it into R, manipulate
 the data, and the functions (and methods) implemented in the package to facilitate easy application of parameters into epidemiological pipelines.
 
+The {distributional} package is loaded as some of the {epiparameter} methods uses S3 generics from this package.
+
 ```{r setup}
 library(epiparameter)
 library(distributional)
--- a/vignettes/epiparameter.Rmd
+++ b/vignettes/epiparameter.Rmd
@@ -14,7 +14,7 @@ knitr::opts_chunk$set(
 )
 ```
 
-It is often the case that in an infectious disease outbreak epidemiological parameters are required in order to characterise and model the dynamics of disease transmission and evaluate control strategies. In those scenarios, epidemiological parameters are commonly retrieved from the literature, and there is no currently library of parameters in order to contrast and compare different reported parameters for a range of infectious diseases and pathogens, from different published studies over time, of which some may be meta-analyses.
+It is often the case that in an infectious disease outbreak epidemiological parameters are required in order to characterise and model the dynamics of disease transmission and evaluate control strategies. In those scenarios, epidemiological parameters are commonly retrieved from the literature, and there is currently no library of parameters in order to contrast and compare different reported parameters for a range of infectious diseases and pathogens, from different published studies over time, of which some may be meta-analyses.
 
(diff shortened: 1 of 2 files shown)


pratikunterwegs (commented):
Thanks for tagging @joshwlambert, Some thoughts are below as well as a few typos caught. I'm going to leave this as a comment, and let you implement which suggestions you think are useful.

On the whole I think this is a good vignette. It does begin with some more advanced concepts (relative to the Readme, say). I think some overlap with the Readme is alright as many users will likely only read this document, so it might be good to have a slightly shallower learning curve.

1. Suggest explaining why you're importing {distributional}.
2. Section levels jump from level 2 (`##`) to level 4 (`####`) in places, suggest cascading levels 2 > 3 > 4 (level 4 if really necessary; suggest the Tufte-ian practice of only having two section levels). Perhaps we should switch to using levels 1 and 2, rather than 2 and 3?
3. `<vb_epidist>` - suggest expanding "vb" and explaining what intrinsic and extrinsic distributions are. Many users will already know, but a one-line explanation might still be useful.
4. *Library of epidemiological parameters* and `<epiparam>` - I think there's going to be some confusion here for new users who aren't used to the fact that the same name is applied to a function that calls the full parameter library, as well as to a class. I imagine that most new users will interact with `<epiparam>` objects mostly to browse the the parameter library. So I would drop the details about how `<epiparam>` has custom print methods etc., and simply say that `epiparam()` returns an object of the class `<epiparam>`, which is very similar to a dataframe, and which can be handled that way. The details - such as demoting `<epiparam>` to a dataframe can be covered in its own section, or in a separate vignette.
	1. I didn't see an explanation of the different types of delay distributions that are available - have I missed that, or is it planned for a different vignette?
	2. I would suggest, following from (1), explaining how the library was populated with a reference to the "Data collation" vignette.
	3. I would also suggest explaining the non-obvious columns; e.g. "truncation", "discretised", "censored", "right_truncated", and "phase_bias_adjusted". The main columns are self-evident, and the parameter columns can be inferred from their names, but these more technical columns might not be familiar to some users and it could be worth adding a short explainer box.
5. *Single set of epidemiological parameters* and `<epidist>` - I would start with a simpler example for `<epidist>` objects, such as the example from the Readme. The Readme might well evolve as the package gains functionality, so some repitition is not a bad thing. The more complex example ("conversion of `<epiparam>` → `<epidist>`") can be explained after a simpler one.
	1. A point that Andree made in the {finalsize} vignettes that might be relevant here - it might be easier to show Tidyverse usage in the complex example. I would suggest showing this as default, as I suspect this is what many users will find more natural (you're already suggesting {dplyr} as it is).
	2. I think most users will want to use `epidist_db()` to create an `<epidist>` from the default library much more than using `as_epidist()` to convert a manually selected row, so it might be worth introducing this function before introducing `as_epidist()`.`
	3. It might be worth providing a table for the distribution name divergence, and explaining why this is the case (something to do with {distributional}? Or was that the symbol names?). Either way, we cannot tell people to Google the answer - they're probably reading this text expecting a comprehensive guide, so it must be more than comprehensive.
6. *Adding library entries* - it might be worth explaining a use case in which might want to do this. It might also be worth explaining that this only adds the entries to the in-environment library, not to, for example, the library in the locally installed version of {epiparameter}. Essentially, that this is not a local config function.
7. *Distribution functions* - I would start with a broader introduction to extracting properties of distributions, and then say something such as, "`<epidist>` objects store the details of distributions, and these properties can be extracted directly from them ..."
8. *Plotting* - I would suggest adding a sentence on using this function for sanity checks on `epidist` objects, such as those that are manually defined; perhaps add some more common use cases?
9. *Parameter conversion* - Again I would suggest a comprehensive list of the available conversions. This should be sufficient as they would then link to their own pages.

joshwlambert:
@pratikunterwegs Thanks for the comments. I agree with most and have made updates in the preceding commits to address.

vignettes/epiparameter.Rmd
```diff
@@ -23,12 +23,14 @@ The {epiparameter} R package is a library of epidemiological parameters, with cl
...
 An outbreak of a known or potentially novel pathogen is detected and the parameters and a delay distribution (e.g. incubation period or serial interval) is required. 
 
-{epiparameter} can provide a these distributions from a selection of published sources in order to provide reliable epidemiological parameters.
+{epiparameter} can provide these distributions from a selection of published sources in order to provide reliable epidemiological parameters.
 :::
 
 This vignette will provide a introduction to the data stored within {epiparameter}, how to read it into R, manipulate
 the data, and the functions (and methods) implemented in the package to facilitate easy application of parameters into epidemiological pipelines.
 
+The {distributional} package is loaded as some of the {epiparameter} methods uses S3 generics from this package.
```
pratikunterwegs, line 32: Typo in "use"

pratikunterwegs (approved):
Happy to approve @joshwlambert, just caught a typo and one small suggestion on which text is linked when referring to blog.

Issue #134, opened 2023-04-17
Inconsistent parameter names when printing


joshwlambert:
Thanks. Although I notice it seems to be converting (mean,dispersion) to (n,p) in epidist object here?

_Originally posted by @adamkucharski in https://github.com/epiverse-trace/epiparameter/pull/128#discussion_r1163968435_

joshwlambert:
Resolving in #137. Closing now.

Issue #98, opened 2023-02-27
Add more distribution conversions


joshwlambert:
Might be nice in the future to extend to more conversions, e.g. could also estimate lognormal parameters using the median instead of the mean.

_Originally posted by @sbfnk in https://github.com/epiverse-trace/epiparameter/pull/85#discussion_r1101340874_

joshwlambert:
If we add more conversions it might make sense to change the setup of the conversion functions. Currently, each function maps one parameterisation to one other, e.g. mean and sd to shape and scale. However, if we want to add more conversions it might make more sense to map one parameterisation to many. E.g. given mean and sd of a gamma distribution return a list of all different parameterisation conversions. 

This would reduce the amount of code and functions, but would make the function names less clear, as currently they are in the style of `params2params()`.

joshwlambert:
The many-to-one and one-to-many conversion functions were introduced in #137. Closing now.

PR #137, opened 2023-04-24
Expands parameter conversion functionality and adds vignette

### LINKED ISSUE #113: Make use of "lognormal" or "lnorm" consistent
Please place an "x" in all the boxes that apply
---------------------------------------------
  
- [x] I have the most recent version of linelist and R
- [x] I have found a bug
- [x] I have a [reproducible example](http://reprex.tidyverse.org/articles/reprex-dos-and-donts.html)
- [x] I want to request a new feature

--------

Currently the use of "lnorm" in `epidist()` is consistent with the current function documentation:

https://github.com/epiverse-trace/epiparameter/blob/ed0da13a8802e364c1e35fff97323e319838a07b/R/epidist.R#L16-L18
  
However, I had a confusing moment when trying to render the epidist example using "lognormal" probability distribution [from the showcase](https://github.com/joshwlambert/epiparameterShowcase/blob/becc6723316d8005bec77e9367f1d9e85b3e0351/vignettes/epiparameter_showcase.Rmd#L85-L96). I also identified that it is mentioned by @joshwlambert as in the TODO list [in calc_dist_params.R](https://github.com/epiverse-trace/epiparameter/blob/ed0da13a8802e364c1e35fff97323e319838a07b/R/calc_dist_params.R#L90-L93) but didn't find it as an issue. My intention here is to record it as a reminder. Here a reprex for this:
  
``` r
library(epiparameter)

epidist(
  disease = "ebola", 
  epi_dist= "incubation_period", 
  prob_distribution = "lnorm", 
  prob_distribution_params = c(meanlog = 1, sdlog = 1)
)
#> Citation cannot be created as either author, year or DOI is missing
#> Disease: ebola
#> Pathogen: NA
#> Epi Distribution: incubation period
#> Study: No citation available
#> Distribution: lognormal
#> Parameters:
#>   mu: 1
#>   sigma: 1

epidist(
  disease = "ebola", 
  epi_dist= "incubation_period", 
  prob_distribution = "lognormal", 
  prob_distribution_params = c(meanlog = 1, sdlog = 1)
)
#> Citation cannot be created as either author, year or DOI is missing
#> parameters class not recognised
#> Error in create_prob_dist(prob_dist = prob_dist, prob_dist_params = prob_dist_params, : Did not recognise distribution name
```

<sup>Created on 2023-03-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

---------

joshwlambert:
This PR addresses:
- #75
- #98
- #134 

The parameter conversion and extraction functionality has been described in a dedicated vignette. This is part of the process of adding topic-specific vignettes #75. The vignette includes case studies from the previous vignette that was removed in #132.

Adds a new style of conversion functions that are more general and accommodates the one-to-many conversion discussed in #98. As part of the new conversion functions, some internal functions have been added (`is_number()`, `chk_ss()` and `get_sd()`).

The `<epidist>` print method now consistently prints parameter names through the use of `family.epidist()` and `parameters.epidist()`, with updates to `clean_epidist_params()`.

Two new exported functions have been added:
- `family.epidist()` 
- `is_truncated()`

These make it easier for users and {epiparameter} developers to check what distribution the `<epidist>` object contains and whether the distribution in the `<epidist>` object is truncated.

This PR also ensure the distribution naming is consistent with the base R style and closes #113.

Two functions have been removed from the package API (i.e. exported -> unexported):
- `extract_param_percentile()`
- `extract_param_range()`

These two functions are called by `extract_param()` and controlled by the `type` argument (either `percentiles` or `range`). There was no need to export these when the functionality can easily be accessed through `extract_param()` and makes it less confusing for users to figure out which function to use.

Tests and documentation are updated inline with the above-mentioned changes.


pratikunterwegs (commented):
Thanks for tagging @joshwlambert, I've taken a look and it seems alright to me overall. I think the three issues indicated in the PR description are addressed.

I've put down some comments, and I think you could merge this after some minor optional edits. The function `is_number()` could go for instance, not sure why it's used in addition to `checkmate::assert_number()`.

1. I have not checked the maths behind the parameter conversion, and I trust that you and the tests have got this right and are keeping it that way.
2. I haven't really understood the `family()` method, but from what I can tell it looks okay.
3. Suggest imperative rather than descriptive function titles, "Add X" instead of "Adds X".

The vignette also seems fine to me - perhaps it could be a bit shorter? The section on conversion especially has a lot of printed lists - maybe one or two examples, and some kind of table or simple text of distribution parameterisations and conversions between them would be sufficient.

Overall, good stuff - I'm putting this down as a 'comment' so Hugo can also take a look if he'd like. I'll be away the rest of the week, so feel free to move ahead as you judge suitable with merging.

R/convert_param.R
```diff
@@ -1,49 +1,301 @@
...
+  # check input
+  chk_ss(x)
+
+  # convert var or cv into sd if available
+  x <- get_sd(x)
+
+  if (is_number(x$mean) && is_number(x$sd)) {
+    # mean and sd to params
+    return(lnorm_meansd2meanlogsdlog(mean = x$mean, sd = x$sd))
+  } else if (is_number(x$median) && is_number(x$sd)) {
```
Bisaloo, line 144:
```suggestion
  }
  if (is_number(x$median) && is_number(x$sd)) {
```

`else` is not needed since you have a `return()` before.

https://speakerdeck.com/jennybc/code-smells-and-feels?slide=45

R/convert_param.R
```diff
@@ -1,49 +1,301 @@
...
+  if (!exists("meanlog") || !exists("sdlog")) {
+    stop("Cannot calculate lognormal parameters from given input")
+  }
+
+  # return list of params
+  list(
+    meanlog = meanlog,
+    sdlog = sdlog
+  )
+}
```
Bisaloo, line 160:
```suggestion
convert_lnorm_summary_stats <- function(...) {

  # capture input
  x <- list(...)

  # check input
  chk_ss(x)

  # convert var or cv into sd if available
  x <- get_sd(x)

  if (is_number(x$mean) && is_number(x$sd)) {
    # mean and sd to params
    return(lnorm_meansd2meanlogsdlog(mean = x$mean, sd = x$sd))
  }
  if (!(is_number(x$median) && is_number(x$sd))) {
    stop("Cannot calculate lognormal parameters from given input")
  }

  # median and sd to params
  sdlog <- sqrt(log(1 + (x$sd / x$median)^2))
  meanlog <- log(x$median) - sdlog^2 / 2

  # return list of params
  list(
    meanlog = meanlog,
    sdlog = sdlog
  )
}
```

To avoid the `exists()` pattern which is a likely source of error if you rename the variables.

R/convert_param.R
```diff
@@ -1,49 +1,301 @@
...
+  # convert var or cv into sd if available
+  x <- get_sd(x)
+
+  if (is_number(x$mean) && is_number(x$sd)) {
+    # mean and sd to params
+    return(gamma_meansd2shapescale(mean = x$mean, sd = x$sd))
+  }
+
+  # if either parameter hasn't been calculated, error
+  if (!exists("shape") || !exists("scale")) {
```
Bisaloo, line 289: Doesn't this always evaluate to FALSE for now? Is this here for future improvements or is it a copy/paste oversight?

R/convert_param.R
```diff
@@ -76,6 +328,106 @@ gamma_meansd2shapescale <- function(mean, sd) {
...
+  checkmate::assert_number(scale, lower = 0)
+
+  # calculate metrics
+  mean <- scale * gamma(1 + 1 / shape)
+  median <- stats::qweibull(p = 0.5, shape = shape, scale = scale)
+  mode <- ifelse(
+    shape > 1,
+    yes = scale * ((shape - 1) / shape)^(1 / shape),
+    no = 0
+  )
```
Bisaloo, line 361:
```suggestion
  mode <- max(scale * ((shape - 1) / shape)^(1 / shape), 0)
```

R/convert_param.R
```diff
@@ -150,6 +501,125 @@ weibull_shapescale2meansd <- function(shape, scale) {
...
+  checkmate::assert_number(dispersion, lower = 0)
+
+  # calculate metrics
+  mean <- dispersion * (1 - prob) / prob
+  median <- stats::qnbinom(p = 0.5, prob = prob, size = dispersion)
+  mode <- ifelse(
+    test = dispersion > 1,
+    yes = (dispersion - 1) * (1 - prob) / prob,
+    no = 0
+  )
```
Bisaloo, line 537:
```suggestion
  mode <- max((dispersion - 1) * (1 - prob) / prob, 0)
```

R/epidist.R
```diff
@@ -814,19 +812,147 @@ parameters.epidist <- function(x, ...) {
...
+    test = prob_dist == "negbin",
+    yes = "nbinom",
+    no = prob_dist
+  )
+
+  prob_dist <- ifelse(
+    test = prob_dist == "geometric",
+    yes = "geom",
+    no = prob_dist
+  )
```
Bisaloo, line 904: Could this be a `switch()` call?

```suggestion
  prob_dist <- switch(
    prob_dist,
    lognormal = "lnorm",
    negbin = "nbinom",
    geometric = "geom",
    prob_dist
  )
```

R/epidist_utils.R
```diff
@@ -581,20 +581,19 @@ clean_epidist_params.gamma <- function(prob_dist_params) {
...
+    names(prob_dist_params)[c(mu_index, sigma_index)] <- c("meanlog", "sdlog")
 
     # remove class attribute from prob_dist_params
     prob_dist_params <- unclass(prob_dist_params)
 
-
     # return prob_dist_params
     return(prob_dist_params)
-  } else if (all(c("mu", "sigma") %in% names(prob_dist_params))) {
+  } else if (all(c("meanlog", "sdlog") %in% names(prob_dist_params))) {
```
Bisaloo, line 596:
```suggestion
  }
  if (all(c("meanlog", "sdlog") %in% names(prob_dist_params))) {
```

R/epidist_utils.R
```diff
@@ -638,7 +637,28 @@ clean_epidist_params.weibull <- function(prob_dist_params) {
...
+
+    # rearrange vector
+    prob_dist_params <- prob_dist_params[c("mean", "dispersion")]
+
+    # remove class attribute from prob_dist_params
+    prob_dist_params <- unclass(prob_dist_params)
+
+    # return prob_dist_params
+    return(prob_dist_params)
+  } else if (all(c("mean", "dispersion") %in% names(prob_dist_params))) {
```
Bisaloo, line 661:
```suggestion
  } 
  if (all(c("mean", "dispersion") %in% names(prob_dist_params))) {
```

vignettes/extract_convert.Rmd
```diff
@@ -0,0 +1,256 @@
...
+:::
+
+```{r setup}
+library(epiparameter)
+```
+
+## Conversions
+
+There are two styles of conversion functions in {epiparameter}. The general conversions
+which have the function name `convert_dist_*()` (where `*` and `dist` are placeholders). These convert
```
Bisaloo, line 50: Wouldn't it be simpler for users to have just the one-to-many conversion and they can then easily extract the relevant parameters to reproduce what the one-to-one are doing?

R/epidist.R
```diff
@@ -814,19 +812,147 @@ parameters.epidist <- function(x, ...) {
...
+#' )
+#' family(edist)
+family.epidist <- function(object, ...) {
+
+  if (inherits(object$prob_dist, "distcrete")) {
+    prob_dist <- object$prob_dist$name
+  } else {
+    prob_dist <- stats::family(object$prob_dist)
+
+    if (identical(prob_dist, "truncated")) {
```
Bisaloo, line 878:
```suggestion
    if (is_truncated(object)) {
```

If I understood correctly

Bisaloo (commented):
I find the new vignette helpful :+1:!

Couple of general comments:

- We should probably aim at keeping the number of exported functions "low" (the exact number depends on the package) to make it easier to understand to new users and easier to maintain. Since you add new exported functions here that offer a superset of functionality of what the 1-to-1 conversions offer, I would recommend removing or unexporting the 1-to-1 conversions.
- It is standard practice to try and put the `stop()` at the start of the function as much as possible. This should also allow you to remove the fragile `exists()` pattern in many (all?) cases
- Similarly, when you have a return() in a branch, it is usually recommended to not put an else after to limit the branching level.

Happy to chat about all of this in a meeting if necessary.

R/convert_param.R
```diff
@@ -1,49 +1,301 @@
...
-#' logarithm) of the distribution
+#' @param x An R object.
+#'
+#' @return Boolean logical
+#' @keywords internal
+#' @noRd
+is_number <- function(x) {
+  is.numeric(x) && length(x) == 1 && !is.na(x)
+}
+
```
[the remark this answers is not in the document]
joshwlambert: Resolved in 5c85d228b3802e52ee11761cd7434f9767161bcf

R/convert_param.R
```diff
@@ -1,49 +1,301 @@
...
+#' Adds standard deviation to the list if not present or errors
+#'
+#' @param x A list of summary statistics
+#'
+#' @return A list of summary statistics
+#' @keywords internal
+#' @noRd
+get_sd <- function(x) {
+  if (isTRUE("var" %in% names(x)) && !isTRUE("sd" %in% names(x)))  {
```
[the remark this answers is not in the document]
joshwlambert: Resolved in f80b384efe516f48a31cab255c2b76d550e0531a

R/convert_param.R
```diff
@@ -1,49 +1,301 @@
...
-#' @param sigma The sigma parameter (standard deviation of the natural
-#' logarithm) of the distribution
+#' @param x An R object.
+#'
+#' @return Boolean logical
+#' @keywords internal
+#' @noRd
+is_number <- function(x) {
+  is.numeric(x) && length(x) == 1 && !is.na(x)
+}
```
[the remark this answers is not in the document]
joshwlambert: Yes, resolved in 5c85d228b3802e52ee11761cd7434f9767161bcf

R/convert_param.R
```diff
@@ -1,49 +1,301 @@
...
+  if (isTRUE("var" %in% names(x)) && !isTRUE("sd" %in% names(x)))  {
+    x$sd <- sqrt(x$var)
+  } else if (all(c("mean", "cv") %in% names(x)) &&
+             !isTRUE("sd" %in% names(x))) {
+    x$sd <- x$cv * x$mean
+  }
+
+  # return list of summary statistics
+  x
+}
```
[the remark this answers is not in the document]
joshwlambert: For some reason the commit suggestion button is not active.

R/convert_param.R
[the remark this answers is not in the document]
joshwlambert: Resolved in e209c59794b4688bb48a2d811136ed26b14b8032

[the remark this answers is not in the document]
joshwlambert: As the function is internal and used only in specific cases I will leave the input checking out.

R/convert_param.R
```diff
@@ -199,6 +667,114 @@ nbinom_meandisp2probdisp <- function(mean, dispersion) {
...
+    "all arguments must be named" =
+      !is.null(names(x)) && isFALSE("" %in% names(x)),
+    "all values given must be numeric" =
+      all(vapply(x, is.numeric, FUN.VALUE = logical(1))),
+    "names of input must match:
+    'mean', 'median', 'mode', 'var', 'sd', 'cv', 'skewness', 'ex_kurtosis'" =
+      all(names(x) %in%  c(
+        "mean", "median", "mode", "var", "sd", "cv", "skewness", "ex_kurtosis"
+      ))
+  )
```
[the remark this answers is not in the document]
joshwlambert: Resolved in a0ff449e92bbfaff0f8f153a62f8258d9850d3f2

R/epidist.R
```diff
@@ -814,19 +812,147 @@ parameters.epidist <- function(x, ...) {
...
+#'   prob_distribution_params = c(meanlog = 1, sdlog = 1),
+#'   truncation = 10
+#' )
+#' is_truncated(edist)
+is_truncated <- function(x) {
+
+  stopifnot(
+    "is_truncated only works for `<epidist> objects`" =
+      inherits(x, "epidist")
+  )
```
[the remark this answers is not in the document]
joshwlambert: Resolved in da9f2521c479adfadc289b01b59aeadcb9dc4659

R/epidist_utils.R
```diff
@@ -581,20 +581,19 @@ clean_epidist_params.gamma <- function(prob_dist_params) {
 clean_epidist_params.lnorm <- function(prob_dist_params) {
 
   # if meanlog and sdlog are provided convert to mu and sigma
```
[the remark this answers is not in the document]
joshwlambert: Resolved in 7a0b9b34b96a6dc37150a4b068b30c40b9900cf5

tests/testthat/test-convert_param.R
```diff
@@ -389,26 +392,335 @@ test_that("geometric_prob2mean works as expected", {
...
 })
+
+test_that("is_number works as expected", {
+  expect_true(is_number(1))
+  expect_false(is_number("1"))
+  expect_false(is_number(c(1, 2)))
+  expect_false(is_number(TRUE))
+  expect_false(is_number(NA))
+  expect_false(is_number(NA_real_))
+})
```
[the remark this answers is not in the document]
joshwlambert: Resolved in 5c85d228b3802e52ee11761cd7434f9767161bcf

vignettes/extract_convert.Rmd
```diff
@@ -0,0 +1,256 @@
...
+or a set of summary statistics input to one set of parameters (e.g. `convert_lnorm_summary_stats()`).
+The other style of function is the one-to-one conversion, which has the function name `dist_param2summary_stat()`
+or `dist_summarystat2param()` (where `dist`, `param` and `summary_stat` are placeholders). These functions take
+two arguments as input and convert to two output (e.g. `gamma_shapescale2meansd()`), with the exception of the geometric conversions which have a single input and a single output.
+
+### General conversion functions
+
+The general conversion functions from summary statistics to parameters (e.g. `convert_gamma_summary_stats()`) have no defined function arguments. Instead the summary statistics should be matched exactly by name.
+
+All currently supported summary statistic conversions in {epiparameter} are given below for each distribution.
```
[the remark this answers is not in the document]
joshwlambert: I've tried to keep the text relatively short. I think that cutting and referring to the function documentation might make it hard for readers and put off newcomers to the package.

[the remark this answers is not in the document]
joshwlambert: I agree that the code chunks are slightly longer than ideal, but as there are not loads of conversions I will leave of them in. Providing a subset raises the question of which to leave out and I think allowing users to see different examples never hurts.

vignettes/extract_convert.Rmd
```diff
@@ -0,0 +1,256 @@
...
+)
+```
+
+The reasoning for the default maximum number of iterations is to limit the computation time and prevent the function cycling through optimisation routines without converging on a consistent answer. If the runtime is not important and parameter accuracy is paramount then the maximum number of iterations can be increased and tolerance decreased. These `control` settings work identically for extracting from percentiles or median and range.
+
+### Use cases
+
+Here we present examples of published epidemiological parameters and distributions where the functions outlined above can be applied to get the parameters of the distribution.
+
+The 75th percentiles reported for a lognormal distribution from @nolen_extended_2016 for the incubation period for mpox (monkeypox).
```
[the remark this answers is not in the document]
joshwlambert: Thanks for pointing this out. I will address this change in an upcoming pull request that works on the formatting of the vignettes and pkgdown site.

vignettes/extract_convert.Rmd
```diff
@@ -0,0 +1,256 @@
...
+  samples = 23
+)
+```
+
+::: {.alert .alert-warning}
+### Assuming distributions
+
+It can be the case that a study will report summary statistics for an unspecified distribution or just for the raw data. In these cases if a parameterised distribution is required for downstream analysis a functional, parametric, form may have to be assumed. 
+
+If the distribution is a delay distribution (i.e. serial interval or incubation period) it can often be sensible to assume a right-skewed distribution such as: gamma, lognormal or weibull distributions. These are also the most commonly fit distributions in epidemiological analysis of delay distributions.
```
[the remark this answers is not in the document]
joshwlambert: Resolved in 1d386dce54db84c110bd96e6e703f307fd80af5f

R/epidist.R
joshwlambert: Thanks for suggesting this change, it's so much nicer than what I'd written!

You win this round of code golf ⛳

vignettes/extract_convert.Rmd
joshwlambert: I'd given this some thought and decided that the one-to-one conversions, although functionally superfluous, might be easier to use given the function name. So I kept them as the cost of maintenance and increasing the package namespace/API would be outweighed by the simplicity and ease of use of having these functions. 

However, saying that, I do not have a strong preference so if you feel the package would be simpler without them, I'm happy to remove and updated to the vignette to explain how to use the one-to-many function in those cases.

R/convert_param.R
joshwlambert: It can evaluate to TRUE if the input summary statistics cannot be used, e.g.:

``` r
epiparameter::convert_gamma_summary_stats(median = 1, sd = 1)
#> Error in epiparameter::convert_gamma_summary_stats(median = 1, sd = 1): Cannot calculate gamma parameters from given input
```

<sup>Created on 2023-04-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

vignettes/extract_convert.Rmd
Bisaloo: Ok, let's park this until #139 is implemented. It should give us a clearer view of how the different exported functions are organized and fit together and help us see what the best way forward is. Does this sound reasonable?

joshwlambert: Agreed.

joshwlambert:
> It is standard practice to try and put the stop() at the start of the function as much as possible. This should also allow you to remove the fragile exists() pattern in many (all?) cases

@Bisaloo would you recommend removing all `exist()` calls? The intended use was to allow the functions to be extendable to new conversion and the `if` statement with the `exist()` would then catch the cases that were not implemented. 

I think using `stop` at the start of functions will require more changes over time as new conversions are added. 

Happy to implement whichever you think is preferable, just let me know.

joshwlambert:
Commit b53aa458a516e16eca4e82741a6a14bd8e1a36f9 removes the use of `exists()`. The logic of these conversion functions is now a series of `if` statements each with an early return, and then the function ends with a `stop()`. Returning an error is the default when none of the `if` statements can executed.

vignettes/extract_convert.Rmd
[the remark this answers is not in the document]
joshwlambert: Referenced this in #140.

R/convert_param.R
joshwlambert: `exists()` were removed in https://github.com/epiverse-trace/epiparameter/commit/b53aa458a516e16eca4e82741a6a14bd8e1a36f9

Bisaloo (approved):

Issue #91, opened 2023-02-07
Update hierarchy of attributes for selecting data entry with `epidist_db()` 


joshwlambert:
Currently, `epidist_db()` selects the study based on largest sample size for a selected disease. However, there should be a more complex hierarchy to decide which aspects of the data are prioritised.

I suggest having the number one criterion being parameter estimates, as currently a study that has a large sample size, but no parameters is preffered.

This hierarchy could also be flexible if specified by a function argument, which would allow the user to decide which are the most important aspects to them.

joshwlambert:
Moving this to a priority item as I don't think there is enough control given to the user for them to handle the entries (rows) in the library when calling this function. This made is difficult to resolve the existing warning that multiple entries were being matched.

PR #160, opened 2023-08-18
Update `epidist_db()` to give more fine-grained control

joshwlambert:
This PR addresses #163 (and #91 which was superseded by #163).

The new implementation uses `subset` argument to allow the use to define subsetting criteria to select the entry from the database they want. It also includes a `single_epidist` argument to allow uses to specify if they only want a single `<epidist>` returned.

It also changes `is_parameterised()` to make it an S3 generic which has methods for `<epiparam>` and `<epidist>` objects.


pratikunterwegs:
Thanks - just flagging that the new implementation would lead to this code

```r
epidist_db(
  disease = "COVID-19",
  epi_dist = "onset_to_death",
  author = "Linton_etal"
)
```

returning multiple `<epidists>` after this change, rather than one, if I'm correct. Will file an issue in {cfr} once this PR is merged, as this would require some code in vignettes and possibly examples to be updated to either use `single_epidist` or manually subset later.

pratikunterwegs (commented):
Looks good to me - discussion continues on Slack so I'd be happy to take another look once the PR stabilises.

R/checkers.R
```diff
@@ -0,0 +1,84 @@
...
+      }
+      if (y$prob_distribution %in% c("geom", "pois")) {
+        out <- !is.na(y$mean)
+        return(out)
+      }
+      return(FALSE)
+    },
+    FUN.VALUE = logical(1),
+    USE.NAMES = FALSE
+  )
```
pratikunterwegs, line 67: This algorithm would return `TRUE` for any comparisons against `Inf` - I'm not sure of the checks on `epiparam()` and whether `Inf` would be allowed at all, but is this a concern?

R/epidist.R
```diff
@@ -333,11 +333,19 @@ validate_epidist <- function(epidist) {
       c("disease", "epi_dist", "prob_dist", "uncertainty", "summary_stats",
         "citation", "metadata", "method_assess", "notes") %in%
       attributes(epidist)$names,
-    "Epidist must contains a disease (single character string)" =
+    "epidist must contains a disease (single character string)" =
```
pratikunterwegs, line 336:
```suggestion
    "epidist must contain a disease (single character string)" =
```

R/epidist_db.R
```diff
@@ -1,22 +1,54 @@
...
+#' [`subset()`]. It is similar to `<data-masking>` using by the `dplyr` package.
+#'
+#' @param single_epidist A boolean `logical` determining whether a single
+#' `<epidist>` or multiple entries from the library can be returned if
+#' matched by the other arguments (`disease`, `epi_dist`, `author`). This
+#' argument is used to prevent multiple sets of parameters being returned
+#' when only one is wanted.
+#'
+#' **Note**: If multiple entries match the arguments supplied and
+#' `single_epidist = TRUE` then the `<epidist>` that is parameterised and
```
pratikunterwegs, line 47: Suggest linking to `is_parameterised()` here for user reference.

R/epidist_db.R
```diff
@@ -1,22 +1,54 @@
...
+#' has the largest sample size will be returned. If multiple entries
```
pratikunterwegs, line 48: What happens when:
1. No entries are parameterised?
2. Only one entry is parameterised, but does not have the largest sample size?

tests/testthat/test-checkers.R
```diff
@@ -0,0 +1,42 @@
+test_that("is_parameterised works as expected for epiparam", {
+  res <- is_parameterised(epiparam())
+  expect_length(res, 118)
```
pratikunterwegs, line 3: Could this be not hardcoded? It will fail when more parameters are added to the package library

tests/testthat/test-checkers.R
```diff
@@ -0,0 +1,42 @@
...
+  expect_type(res, type = "logical")
+})
+
+test_that("is_parameterised works as expected with epiparam parameters", {
+  expect_true(is_parameterised(epiparam()[15, ]))
```
pratikunterwegs, line 8: Again suggest not hardcoding examples as their position might change

tests/testthat/test-epidist_db.R
```diff
@@ -34,21 +21,66 @@ test_that("epidist_db works as expected with author specified", {
...
+  edist <- suppressMessages(
+    epidist_db(
+      disease = "influenza",
+      epi_dist = "incubation_period",
+      subset = sample_size == max(sample_size)
+    )
+  )
+
+  expect_s3_class(edist, class = "epidist")
+  expect_length(edist, 9)
```
pratikunterwegs, line 38: Why this expectation on `edist` length? The number of class members of `<epidist>` should be constant, yes?

tests/testthat/test-epidist_db.R
```diff
@@ -34,21 +21,66 @@ test_that("epidist_db works as expected with author specified", {
...
+  expect_s3_class(validate_epidist(edist), class = "epidist")
```
pratikunterwegs, line 39: Also curious about this use of the validator here - would this not be applied already in `epidist_db()`?

tests/testthat/test-epidist_db.R
```diff
@@ -34,21 +21,66 @@ test_that("epidist_db works as expected with author specified", {
...
+  edist <- suppressMessages(
+    epidist_db(
+      disease = "COVID-19",
+      epi_dist = "incubation_period",
+      subset = is_parameterised
+    )
+  )
+
+  expect_type(edist, "list")
+  expect_length(edist, 2)
```
pratikunterwegs, line 52: (here and elsewhere): Hardcoded - might be brittle to database expansion

pratikunterwegs (commented):
Thanks for making these changes @joshwlambert - looks okay to me other than some minor points flagged here.

R/checkers.R
joshwlambert: It is not likely that the parameter values or summary statistics will in `Inf`. I will make an issue and then see if this step can be checked in the JSON validation stage to prevent this issue occurring interactively,

R/epidist_db.R
pratikunterwegs: Okay thanks - worth adding to the docs in a couple of lines? Ignore if already explained in other words

tests/testthat/test-checkers.R
joshwlambert: Its not possible to remove all hardcoding as updating the library may break the test. I've reduced the use of magic numbers in 240c42f882ed9063c87b3e2b00f5f20e1cda0159.

tests/testthat/test-checkers.R
joshwlambert: Resolved in 240c42f882ed9063c87b3e2b00f5f20e1cda0159 (see above).

tests/testthat/test-epidist_db.R
joshwlambert: I've now moved the validation from the tests into the `epidist_db()` function in 0ad700bc6a040c92f3489e40b493cc540780c0a4.

R/epidist_db.R
joshwlambert: Resolved in 0ad700bc6a040c92f3489e40b493cc540780c0a4.

R/epidist.R
joshwlambert: Resolved in 0ad700bc6a040c92f3489e40b493cc540780c0a4.

pratikunterwegs (approved):
Looks good to me, thanks for making the changes. Just the one issue of unused `eparam` in `test-checkers.R`, but otherwise feel free to merge.

R/checkers.R
```diff
@@ -0,0 +1,84 @@
...
+#' is missing either a probability distribution or parameters for the
+#' probability distribution, `is_parameterised` returns `FALSE`, otherwise
+#' it returns `TRUE`.
+#'
+#' @param x An `epidist` or `epiparam` object.
+#' @param ... [`dots`] not used, extra arguments supplied will cause a warning.
+#'
+#' @return A boolean logical for <epidist> or vector of boolean for each entry
+#' in the <epiparam>.
+#' @export
```
Bisaloo, line 14: Since this function is exported, would it be possible to add in the description or in the examples a typical use case for users? At the moment, I can only think about internal use cases.

PR #187, opened 2023-09-11
Add `mean()` method for `<epidist>`

joshwlambert:
This PR adds a mean S3 method for the `<epidist>` class (`mean.epidist()`). This functionality was requested from discussions around how to easily access the mean of a distribution, which is nested with the summary statistics in the `<epidist>` object, and what to do if it is not provided. 

This `mean()` method returns the mean if given, will attempt to calculate it from parameters if not given, and if neither are possible will return `NA`.


pratikunterwegs (commented):
Thanks @joshwlambert - love this! Solves the faff around accessing summary stats for {cfr} really nicely. I see a warning on an `if-else` ladder within the method, so perhaps something to fix? Otherwise happy to see this merged.

joshwlambert:
From what I can see the lintr workflow is failing due to a different function, and it is identifying an `if-else` which isn't there. Can you check and see whether you agree if this is a false positive from {lintr}?

pratikunterwegs:
Running lintr locally I get warnings from four other locations, but not for the if-else ladder identified here.

`test-convert_params.R`: 117, 175; duplicate arguments

`test-vb_epidist.R`: 275, 304; implicit assignment

PR #192, opened 2023-09-22
Restore normal vdiffr workflow

Bisaloo:
Since 1.0.7 should fix the issues that led to version pinning.


joshwlambert:
Thanks for the PR @Bisaloo!

PR #194, opened 2023-09-27
Update epidemiological parameter library

joshwlambert:
This PR updates the library of epidemiological parameters included in the {epiparameter} package. Changes include:

- The library is converted from a CSV format to a JSON format. This change was made to include all authors in each entrie's citation information to produce a complete `<bibentry>` object in R
- The data dictionary (schema) was converted from YAML to JSON
- The validate-json.yaml workflow is simplified by using a JSON database and schema, so removes the need for conversions
- {jsonlite} is taken on a package dependency
- Updates to `epiparam()`, `epidist_db()` and `create_epidist_citation()` are made to work with the updates to the database

Tests and documentation are updated in line with changes made.


tests/testthat/_snaps/epiparam.md
```diff
@@ -8,15 +8,15 @@
...
-      Number of studies in library: 57
+      Number of studies in library: 45
       <Head of library>
             disease         epi_distribution prob_distribution
       1  Adenovirus        incubation_period             lnorm
       2 Chikungunya        incubation_period             lnorm
       3    COVID-19 hospitalisation_to_death           weibull
       4    COVID-19 hospitalisation_to_death             lnorm
       5    COVID-19        incubation_period              <NA>
-      6    COVID-19        incubation_period           weibull
```
pratikunterwegs, line 19: This entry has changed from Weibull to `NA` but checking locally shows it's an issue of row reordering.

R/epiparam.R
pratikunterwegs, line 1: Suggest adding an example to the `epiparam()` documentation showing how to subset by author, as how authors are listed has changed. This can be done with `epidist_db()`, but the return type is then a list of `<epidist>`s.

pratikunterwegs (commented):
Thanks @joshwlambert, I think moving the package in this direction with the parameters held as JSON entries is a good idea overall. Looks alright to me, and could be merged once checks pass.

R/epiparam.R
joshwlambert: Given planned development on `epiparam()` I won't make this change in this PR. I'll log that it should be clearly documented to the user how to subset the library of parameters by author in an issue to make sure future developments bear this in mind.

joshwlambert: #195

tests/testthat/_snaps/epiparam.md
joshwlambert: Yes, I've also checked this and the data is the same it is just the ordering that changed. I think this might be due to the author information changing: https://github.com/epiverse-trace/epiparameter/blob/main/R/epiparam.R#L53-L58

PR #220, opened 2023-12-06
Add internal function to `<epidist>` attributes to load {epiparameter}

joshwlambert:
This PR adds an internal {epiparameter} function, `.epiparameter_namespace()`, to the `<epidist>` constructor. This change ensures that {epiparameter} is _loaded_ (not attached) when using `<epidist>` objects. This covers a scenario where a user may save an `<epidist>` and use it in another R session where {epiparameter} is not installed or loaded.

A more comprehensive explanation of the problem and the mechanism of the solution can be found in this great blog post [Helping R find the right methods for unserialized complex objects by Mike Mahoney](https://www.mm218.dev/posts/2023-11-27-objects-loading-namespaces/) (thanks to @Bisaloo for pointing out the post).


tests/testthat/test-epidist_db.R
```diff
@@ -168,6 +168,7 @@ test_that(".read_epidist_db works as expected", {
 })
 
 test_that("sysdata is the same as .read_epidist_db output", {
+  skip_on_covr()
```
Bisaloo, line 171: Can you add a comment explaining why this is necessary so we don't forget please?

joshwlambert: Resolved in [8ba42c0](https://github.com/epiverse-trace/epiparameter/pull/220/commits/8ba42c0c745d33cca11424c7d3f3070ae75ed9f5). 

I cannot provide a more detailed comment as I could not get to the bottom of what was causing the error. I used `skip_on_covr()` as a fallback for now until I resolve the problem. The error message is relatively uninformative and reading through the {covr} source code it was not clear which point was causing the error (likely [here](https://github.com/r-lib/covr/blob/54de614f5c17a42bfaef2c87dbf5723ca701326a/R/replace.R#L12)).

Bisaloo: covr works by modifying the functions in your package on-the-fly and roughly speaking adding `covr:::count()` calls to get the number of hits on each line.

So if your tests end up checking the body of a function in the tests, you're going to see this kind of errors.

PR #221, opened 2023-12-11
Update documentation for `accessor` S3 generics and methods

Bisaloo:
The generic docs can properly be improved. Feel free to push directly to this branch if you'd like any additional change before merging.


joshwlambert:
@Bisaloo thanks for the changes. I've also used the same documentation structure for the `get_citation()` generic and methods within this PR.

I think this is a relatively easy mistake to make with documenting S3 functions, would it be worth me drafting something for _blueprints_ to guide others in Epiverse-TRACE?

Bisaloo:
> I think this is a relatively easy mistake to make with documenting S3 functions, would it be worth me drafting something for blueprints to guide others in Epiverse-TRACE?

It's probably out of scope for the blueprints since it's not something special we're doing in our community or a general guiding principle.
It could be a blog post. Especially if the scope is extended a bit. E.g., "Documenting S3 generics and methods".
Or just a small nugget of knowledge shared on slack and/or twitter.

Issue #225, opened 2023-12-12
how to subset `multi_epidist` object from `epidist_db()` by the `prob_distribution`?


joshwlambert:
### Discussed in https://github.com/epiverse-trace/epiparameter/discussions/216

<div type='discussions-op-text'>

<sup>Originally posted by **avallecam** November 30, 2023</sup>
I want to extract the distribution for COVID-19 serial interval from Nishiura. If I use the `single_epidist = T` I get the `lnorm`. If I want the `weibull`, how can I do it? or is there a reason to prefer one over the other? or, how should I write the subset?

``` r
library(epiparameter)
library(tidyverse)

# from code below, I identify that prob_distribution column can be used to subset

epiparameter::epidist_db(
  disease = "covid",
  epi_dist = "serial",
  author = "Nishiura") %>%
  epiparameter::list_distributions()
#> Returning 2 results that match the criteria (2 are parameterised). 
#> Use subset to filter by entry variables or single_epidist to return a single entry. 
#> To retrieve the short citation for each use the 'get_citation' function
#>    disease epi_distribution prob_distribution       author year
#> 1 COVID-19  serial interval             lnorm Hiroshi .... 2020
#> 2 COVID-19  serial interval           weibull Hiroshi .... 2020

# I use subset, but this notation does not work

epiparameter::epidist_db(
  disease = "covid",
  epi_dist = "serial",
  author = "Nishiura",
  subset = prob_distribution == "lnorm")
#> Error in epiparameter::epidist_db(disease = "covid", epi_dist = "serial", : object 'prob_distribution' not found

# I get lnorm with single_epidist = T
# if interested in weibull, how to do it?

epiparameter::epidist_db(disease = "covid",
                         epi_dist = "serial",
                         author = "Nishiura",
                         single_epidist = T)
#> Using Nishiura H, Linton N, Akhmetzhanov A (2020). "Serial interval of novel
#> coronavirus (COVID-19) infections." _International Journal of
#> Infectious Diseases_. doi:10.1016/j.ijid.2020.02.060
#> <https://doi.org/10.1016/j.ijid.2020.02.060>.. 
#> To retrieve the short citation use the 'get_citation' function
#> Disease: COVID-19
#> Pathogen: SARS-CoV-2
#> Epi Distribution: serial interval
#> Study: Nishiura H, Linton N, Akhmetzhanov A (2020). "Serial interval of novel
#> coronavirus (COVID-19) infections." _International Journal of
[...]

Issue #224, opened 2023-12-12
how to know which variables are available to subset by in a `multi_epidist` object?


joshwlambert:
### Discussed in https://github.com/epiverse-trace/epiparameter/discussions/217

<div type='discussions-op-text'>

<sup>Originally posted by **avallecam** November 30, 2023</sup>
From the examples, we can filter by `sample_size` (code 1). In a `multi_epidist` object from `epidist_db()` of two elements (code 2), I tried to apply that filter but `object 'sample_size' not found` (code 3). 

Is there a way to see what variables are available for a `multi_epidist` object?

``` r

# code 1

epiparameter::epidist_db(
  disease = "SARS",
  epi_dist = "offspring_distribution",
  subset = sample_size > 40
)
#> Returning 1 results that match the criteria (1 are parameterised). 
#> Use subset to filter by entry variables or single_epidist to return a single entry. 
#> To retrieve the short citation for each use the 'get_citation' function
#> Disease: SARS
#> Pathogen: SARS-Cov-1
#> Epi Distribution: offspring distribution
#> Study: Lloyd-Smith J, Schreiber S, Kopp P, Getz W (2005). "Superspreading and
#> the effect of individual variation on disease emergence." _Nature_.
#> doi:10.1038/nature04153 <https://doi.org/10.1038/nature04153>.
#> Distribution: nbinom
#> Parameters:
#>   mean: 1.630
#>   dispersion: 0.160

# code 2

epiparameter::epidist_db(
  disease = "covid",
  epi_dist = "incubation",
  author = "McAloon"
)
#> Returning 2 results that match the criteria (2 are parameterised). 
#> Use subset to filter by entry variables or single_epidist to return a single entry. 
#> To retrieve the short citation for each use the 'get_citation' function
#> [[1]]
#> Disease: COVID-19
#> Pathogen: SARS-CoV-2
#> Epi Distribution: incubation period
#> Study: McAloon C, Collins Á, Hunt K, Barber A, Byrne A, Butler F, Casey M,
#> Griffin J, Lane E, McEvoy D, Wall P, Green M, O'Grady L, More S (2020).
#> "Incubation period of COVID-19: a rapid systematic review and
#> meta-analysis of observational research." _BMJ Open_.
#> doi:10.1136/bmjopen-2020-039652
#> <https://doi.org/10.1136/bmjopen-2020-039652>.
#> Distribution: lnorm
#> Parameters:
#>   meanlog: 1.660
[...]

PR #226, opened 2023-12-12
Update subsetting in `epidist_db()`

joshwlambert:
This PR addresses #224 & #225. 

It uses the brackets (`[]`) to subset instead of using `subset()`, as this was incorrectly evaluating the expression passed to the `subset` argument in `epidist_db()` when `author` and `subset` were both given. This follows advice from `subset()` documentation stating:

> [This is a convenience function intended for use interactively. For programming it is better to use the standard subsetting functions like [], and in particular the non-standard evaluation of argument subset can have unanticipated consequences.](https://github.com/wch/r-source/blob/trunk/src/library/base/man/subset.Rd#L66-L71)

It also adds a special case to `.is_cond_epidist()` to subset by `prob_dist`.

### DIFF
--- a/R/epidist_db.R
+++ b/R/epidist_db.R
@@ -507,15 +507,25 @@ epidist_db <- function(disease = "all",
 .is_cond_epidist <- function(lst, condition, nse_subject) {
   # check input
   stopifnot(is_epidist(lst))
-  # apply condition to <epidist> name matching elements
-  lapply(lst, function(x) {
-    if (nse_subject %in% names(x)) {
-      # <epidist> is only nested once so no need for recursive search
-      eval(expr = condition, envir = x)
+  if (nse_subject == "prob_dist") {
+    # special case to subset by dist as name is extracted first
+    if (is.object(lst$prob_dist) || is.character(lst$prob_dist)) {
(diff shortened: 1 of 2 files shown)


Bisaloo:
This fixes the issue at hand but I'm not sure about the direction this is going.

You mentioned that filtering by, e.g., date of study, would have to be special cased as well since this piece of info is nested in an bibentry object.

Additionally, as discussed in #224, it is not clear which variables can be used in `subset`.

Despite my initial thoughts when you brought up this issue, I now think that it's better to remove `subset` and provide explicit arguments to subset instead, as is done for `disease`, `epi_dist` and `author`.
However, with this approach, it's probably important to be quite opinionated and accept the fact that no out of the box mechanism will be provided for rare subset conditions. Instead, `epidist_db()` focuses on 4-5 common subsetting parameters.

I'm still not super happy about this solution but it seems like an unavoidable consequence of the complex structure of `(multi_)epidist` objects :thinking: 

As discussed, I do believe it's one of the cases where already having an API hosting the database may help us think more clearly about the best design.

joshwlambert:
@Bisaloo given that these changes fix the issues linked to in the PR description should I merge this and then plan a more general solution, with the possible use of an API for an upcoming PR?

PR #272, opened 2024-03-28
`get_citation.multi_epidist()` return a `<bibentry>`

joshwlambert:
This PR updates the `get_citation()` `<multi_epidist>` S3 method to return a `<bibentry>` object which can contain multiple references. This was requested in #259 and allows multiple references to use the `<bibentry>` methods, instead of having to loop or `lapply` over the list of references.

### DIFF
--- a/.lintr
+++ b/.lintr
@@ -9,7 +9,9 @@ linters: all_linters(
       modify_defaults(
         default_undesirable_functions,
         citEntry = "use the more modern bibentry() function",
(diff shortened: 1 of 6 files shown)


joshwlambert:
Merging PR even though R CMD Check from Ubuntu R devel is failing to build the `extract_convert.Rmd` vignette. This is unrelated to the changes made in this PR and likely due to changes in R in the devel version.

PR #284, opened 2024-04-26
Remove `@examples` in `calc_dist_params()`

joshwlambert:
This PR removes the `@examples` documentation from `calc_dist_params()`. The `calc_dist_params()` function had an example that was erroring but was being missed as it was wrapped in a `\dontrun{}`. This was pointed out in #259. I have removed the example documentation from this function as the function is internal, and only exported functions in this package should have example documentation. `calc_dist_params()` has tests to ensure the function is working as intended.

### DIFF
--- a/R/calc_dist_params.R
+++ b/R/calc_dist_params.R
@@ -29,24 +29,6 @@
 #'
 #' @return A named `numeric` vector with parameters.
 #' @keywords internal
-#' @examples
-#' \dontrun{
-#' calc_dist_params(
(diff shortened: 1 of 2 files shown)


joshwlambert:
The R CMD check GitHub actions workflow is failing due to an issue building the `extract-bias.Rmd` vignette. This CI failure is unrelated to the change in this PR and therefore I will merge and tackle the failing CI in a separate PR.

PR #285, opened 2024-04-29
Try to get more info on failure

Bisaloo:


joshwlambert:
@Bisaloo shall we close this PR? 

We can leave the branch for a couple of weeks to try other solutions if we think of them and then it can be deleted.

PR #286, opened 2024-05-01
Precompile `extract-bias.Rmd` vignette

joshwlambert:
This PR addresses the issue that the `extract-bias.Rmd` vignette has been failing the R CMD check on continuous integration from around the release of R 4.4.0. The error found on CI cannot be replicated locally across different OSs. This PR precompiles this vignette and thus it is not rendered by the R CMD check when run on CI.


.Rbuildignore
```diff
@@ -14,3 +14,4 @@
 ^CITATION\.cff$
 ^data-raw$
 ^tools$
+^vignettes/figure$
```
Bisaloo, line 17: I don't think that's quite right, we want these figures to be included. OTOH, the `.orig` file should be ignored.

joshwlambert: When `^vignettes/figure$ is removed from `.Rbuildignore` the R CMD check returns a note asking if the folder is left over from knitr.

Bisaloo: You can change the `fig.path` value in `knitr::opts_chunk$set()` to save figs directly in the `vignettes/` folder, without subfolders.

joshwlambert: When I set the `fig.path = "vignettes/figure/" it saves the figures in the correct place, but by specifying the path, the path in the vignette to render the images changes to "vignettes/figure/" which means they cannot locate files.

Issue #264, opened 2024-03-26
`as_epidist()` S3 generic to enable conversion of other epidemiological parameter databases


joshwlambert:
To enable an easy, extendable way to convert between other formats that hold epidemiological parameters (e.g. {epireview}), to `<epidist>` objects for applications.

joshwlambert:
Considerations for epidemiological parameters that use base R classes (e.g. vectors or `<data.frames>`) for S3 dispatch, might need to assign classes based on columns or identifiable aspects of data.

PR #298, opened 2024-05-16
Convert {epireview} data into `<epidist>`

joshwlambert:
This PR addresses #264 and adds a **work-in-progress** feature to convert between entries in the {epireview} R package epidemiological parameter tables to `<epidist>` objects for use in pipelines. The `as_epidist()` S3 generic is added to the package, with an `as_epidist.data.frame()` method for converting from {epireview}.

This feature is not fully tested or documented, but is being merged into `main` to be tested in the WHO EpiParameter workshop for user testing. A future PR will ensure the coercion functions are robust and tested to a more diverse range of input.

A new data object `epireview_core_cols` is added to the package to identify which `<data.frame>`s are from {epireview}. The `data-raw` script is also added to the package.

This PR adds a vignette that describes the `as_epidist()` methods with various use cases. 

The existing coercion methods `as.function.epidist()` and `as.data.frame.epidist()` are moved to the new `coercion.R` file.


joshwlambert:
The `lintr` GHA workflow is failing but I will address this in a follow-up PR.

Issue #263, opened 2024-03-26
{epiparameter} Shiny App or Quarto Dashboard


joshwlambert:
A dashboard to explore the parameter library and easier visualise the distributions would be a useful supplementary tool.

jamesmbaazam:
I took a quick stab at this yesterday. It currently lives in my personal repo and at an rudimentary stage at the moment. Will link the raw version by close of today when I publish it.

jamesmbaazam:
This has now been pushed to https://github.com/jamesmbaazam/epiparameterDashboard. You can access the deployed app at https://james-mba-azam.shinyapps.io/epiparameterdashboard/.

joshwlambert:
As there is now a repository for the {epiparameter} dashboard I'll close this issue and issues can be opened in the other repository: https://github.com/epiverse-trace/epiparameterDashboard/issues.

PR #311, opened 2024-05-17
add database vignette

### LINKED ISSUE #296: Database vignette
I wonder if it would be nice to have a vignette that shows all the entries in the database - so that people can check if there are useful distributions in there for them without having to install the package.

sbfnk:
Closes #296.

* **Please check if the PR fulfills these requirements**

- [X] I have read the CONTRIBUTING guidelines
- [ ] A new item has been added to `NEWS.md`
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] Checks have been run locally and pass

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

It adds a prototype for a database vignette

* **Other information**:

Format etc. can be discussed as well as whether this is desirable to have. Creating this PR as a suggestion for what the database might look like.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -11,6 +11,8 @@ Authors@R: c(
            comment = c(ORCID = "0000-0003-4184-2864")),
     person("Hugo", "Gruson", , "hugo.gruson@data.org", role = c("ctb", "rev"),
            comment = c(ORCID = "0000-0002-4094-1476")),
+    person("Sebastian", "Funk", , "sebastian.funk@lshtm.ac.uk", role = "ctb",
+           comment = c(ORCID = "0000-0002-2842-3406")),
     person("Pratik", "Gupte", , "pratik.gupte@lshtm.ac.uk", role = "rev",
            comment = c(ORCID = "0000-0001-5294-7819"))
   )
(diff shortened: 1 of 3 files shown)


joshwlambert:
Thanks @sbfnk this looks great. I'm happy to move forward with this change and merge into `main`. A couple of points before merging.

1. When I checked that the pkgdown site would render it correctly (`pkgdown::build_site()`), the table was not showing. I believe this is because {DT} tables are HTML widgets which requires `pkgdown as_is: true` to be removed from the vignette header. Are you finished pushing to this branch? If so I'll make this change before merging to make sure the vignette renders correctly on the website.

2. In your original issue you said

> so that people can check if there are useful distributions in there for them without having to install the package.

If we want this to be only on the website and not shipped with the package we could consider converting this from a [vignette to an article](https://r-pkgs.org/website.html#vignettes-and-articles). We can always change this down the road so happy to merge before we've decided on this point.

sbfnk:
1. Yes, please go ahead.
2. That sounds like a good idea to me - presumably e.g. on CRAN the table wouldn't be rendered either

jamesmbaazam:
This is a nice addition and might be worth evaluating its benefit over the suggested dashboard in #263 or what gap the dashboard would fill that this database doesn't.

joshwlambert:
> presumably e.g. on CRAN the table wouldn't be rendered either

I wasn't sure whether it would render on CRAN, according to this page I think it does: https://cran.r-project.org/web/packages/DT/vignettes/DT.html. I'll leave as a vignette for now and if we think it is better for this to only be on the website and not on CRAN we can convert to an article in a subsequent PR.

> This is a nice addition and might be worth evaluating its benefit over the suggested dashboard in https://github.com/epiverse-trace/epiparameter/issues/263 or what gap the dashboard would fill that this database doesn't.

I see these two things as _positive duplication_. There will be some overlap between this vignette and the dashboard, but 1) it doesn't hurt to provide the ability to search through the database in multiple places, 2) the dashboard is still in development and this provides a quick solution, 3) the dashboard will provide more functionality that just searching through the database.

joshwlambert:
Apologies for the spam. It took a few commits to get the lintr workflow to pass. 

This PR is now ready to be merged.

PR #314, opened 2024-05-23
Ensure date is `<Date>` in `CITATION` file

### LINKED ISSUE #297: `inst/CITATION` will fail CRAN submission
An issue that we just came across in EpiNow2: the `inst/CITATION` file refers to `meta$Date` (which would be read from the `DESCRIPTION` file) which, if NULL, is replaced with the system date.

This will be an issue upon CRAN submission if `Date` is not set in `DESCRIPTION`. In that case CRAN automatically gets populated with a POSIX format date so it won't be NULL but the `format` command will produce a warning.

The solution, I think, is to just use the system date.

joshwlambert:
This PR closes #297.

### DIFF
--- a/inst/CITATION
+++ b/inst/CITATION
@@ -11,7 +11,7 @@ if (length(author)) {
 if (is.null(meta$Date)) {
   date <- format(Sys.Date(), "%Y")
 } else {
-  date <- format(meta$Date, "%Y")
+  date <- format(as.Date(meta$Date), "%Y")
 }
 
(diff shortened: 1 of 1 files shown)


Bisaloo:
Note that linelist also includes another tweak to the URL field to ensure only one URL is provided

joshwlambert:
Thanks for flagging, if this causes an issue upon CRAN submission I'll update it to follow {linelist}.

PR #329, opened 2024-06-06
Fix pdf manual generation

Bisaloo:
This is an attempt to fix the error reported by r-universe during the pdf manual generation: https://github.com/r-universe/epiverse-trace/actions/runs/9271064360/job/25505583035


joshwlambert (approved):
Thanks for contributing this patch!

PR #330, opened 2024-06-07
Remove some unnecessary nolint comments

Bisaloo:
I understand https://github.com/epiverse-trace/epiparameter/pull/330/commits/5bc086f8122e5323c7aa352ddff92337e33788d7 was flagging an actual lintr bug but I find the `switch()` + `stopifnot()` syntax very difficult to read so I wonder if that's worth bothering with it. Feel free to exclude this commit if you feel strongly.

### DIFF
--- a/R/coercion.R
+++ b/R/coercion.R
@@ -397,9 +397,10 @@ epireview_to_epidist <- function(x, ...) {
 
   # clean parameter type name by removing the parameter class
   epi_dist <- gsub(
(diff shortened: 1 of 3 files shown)


joshwlambert (approved):
Thanks for the updates @Bisaloo! 

I don't have a preference of `switch()` or `if` with `stopifnot()` so happy to make your suggested change.

PR #337, opened 2024-06-26
Add data license to package

### LINKED ISSUE #300: Data license?
Currently **epiparameter** provides original data, but that data does not carry a license. MIT license covers the code, but in general it is not recommended to do this for data as it is not intended for that (just like CC licenses are not intended for code).

I don't want to discuss about whether data can be copyrighted to begin with (I can if desired 😊 ); I do think that not having a license can make it more uncertain about how it can be reused freely.

I would recommend to use a separate license for the data, specifically a Public Domain Dedication. This is maximally permissive and also the standard for (meta)data that is aimed to be reused and recompiled. CC BY becomes problematic when compiling databases, as keeping track of the origins of all data points quickly overtakes the size of the actual data (especially problematic in larger databases).

I also mentioned this in the Collaboratory forum for the GREP database: https://collab-forum.who.int/t/epiparameter-hackathon-discussion-thread/186/2?u=chartgerink

joshwlambert:
This PR addresses #300 by adding the [CC0](https://creativecommons.org/public-domain/cc0/) license to the `LICENSE` file. This resolves #300 and should be a CRAN-compliant way to license both the code and the data in the package in an unambiguous way. See #300 for discussion and considerations on which license was chosen and how it should be included in the package.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 YEAR: 2024
 COPYRIGHT HOLDER: epiparameter authors
-
-All data included in the epiparameter R package is licensed under [CC0](https://creativecommons.org/publicdomain/zero/1.0/legalcode.txt). This includes the parameter database (extdata/parameters.json) and data in the data/ folder.
-
-Please cite the individual parameter entries in the database when used.
+DATA LICENSE: All data included in the epiparameter R package is licensed under CC0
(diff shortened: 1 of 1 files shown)


joshwlambert:
`LICENSE` written using `write.dcf()` to try and make file a valid DCF to fix R CMD check workflow.

PR #346, opened 2024-07-02
Add distribution parameters to Guzzetta et al. Mpox incubation period

### LINKED ISSUE #345: mpox incubation period discrepancy
Please place an "x" in all the boxes that apply
---------------------------------------------

 - [X] I have the most recent version of this package and R
 - [X] I have found a bug
 - [X] I have a [reproducible example](http://reprex.tidyverse.org/articles/reprex-dos-and-donts.html)
 - [ ] I want to request a new feature

--------

The following code (which was going to use as [example for homepage](https://github.com/epiverse-trace/epiverse-trace.github.io/issues/292)) produces output parameters as below
```r
epidist_db(disease="mpox", epi_dist="incubation_period", single_epidist = T)
```
Parameters:
  shape: 2.437
  scale: 3.678

But in the original paper appendix, the shape is given as 2.42 (1.26–3.62) and scale by 3.75 (2.17–7.13). Just wondered what was driving this (small) difference?

joshwlambert:
This PR fixes #345 by adding the Gamma distribution parameters reported in the appendix of the [Guzzetta et al. (2022) paper](https://wwwnc.cdc.gov/eid/article/28/10/22-1126_article). The sample size and transmission mode in the database entry have also been fixed.


adamkucharski (approved):
Thanks, that is working well now. And good spot on the sample size

Issue #325, opened 2024-05-29
Potential naming conflict


sbfnk:
Just flaggin that the `epidist` class vs. `epidist` package might cause confusion in the future
https://github.com/epinowcast/epidist

joshwlambert:
Thanks for flagging, I had noticed this but didn't feel like it was likely to cause many issues. It does emphasise the notation and clarity of discussing packages, functions, and classes. 

At the moment, my preference is not to change anything. If we decided the class name should change then `<epiparameter>` could be a suitable name as many classes are named after the package that exports them (e.g. `<incidence>`, `<epicontacts>`).

joshwlambert:
Tagging @jamesmbaazam as he mentioned the same thing recently.

jamesmbaazam:
Yes, I was wondering if it would be worth discussing and fixing sooner rather than later. `<epidistribution>` or `<epiparameter>` are good candidates.

joshwlambert:
I don't think anything needs fixing, as it assumes something is broken. There is no technical conflict (e.g. namespace conflict) as the {epidist} package doesn't contain an `epidist()` function (https://epidist.epinowcast.org/reference/index.html). 

Although changing the class names in {epiparameter} could be done quickly we don't want to break scripts using the package. Hence, why I've been hesitant to make this change straight away.

I think the best next steps are to contact the maintainer(s) of {epidist} and discuss what would be best for both packages. What do you both think?

joshwlambert:
The `epidist` class will be renamed `epiparameter` in v0.3.0.

PR #360, opened 2024-07-17
Rename `<epidist>` with `<epiparameter>`

joshwlambert:
***Breaking changes***

This PR addresses #325 by renaming the main class in {epiparameter} from `<epidist>` to `<epiparameter>`. This is due to potential conflict or confusion with the [{epidist} R package](https://github.com/epinowcast/epidist).

Function names that previously used `epidist` now use `epiparameter` (for exceptions see below). These include: `epiparameter_db()`, `epiparameter()`, `as_epiparameter()`, etc. The other class used in {epiparameter}, `<multi_epidist>` has also been renamed to `<multi_epiparameter>`.

Naming the class after the package also follows the idiomatic convention of other R packages, e.g. {incidence} and `<incidence>`, {epicontacts} and `<epicontacts>`.

The other change made in this PR is to shorten the names of the `create_*()` and `.clean_*()` functions by removing `epidist` from the name. If there is a conflict with other functions with the same name from another package then the functions can be explicitly namespaced (`epiparameter::create_uncertainty()`).


jamesmbaazam (commented):
This renaming exercise looks good to me. My only concern is that this package might already have a good number of users who will be affected by this change without prior warning. 

A better approach would be to [gradually deprecate the functions](https://lifecycle.r-lib.org/articles/communicate.html#gradual-deprecation). I'm sure you have good reasons for going this route but I just thought to bring it up for discussion.

joshwlambert:
@jamesmbaazam I appreciate this being raised. In general for breaking changes on packages that are < v1.0.0 and are not on CRAN I've made breaking changes without a gradual deprecation process. I always record any breaking changes in the `NEWS.md` file.

Issue #364, opened 2024-08-12
Unify S3 class design principles with {contactmatrix}


joshwlambert:
The [{contactmatrix}](https://github.com/socialcontactdata/contactmatrix) R package aims to unify the underlying data structure used by packages that use contact matrices. {epiparameter} aims to do some very similar with its `<epiparameter>` class (previously `<epidist>`). Therefore, as these two projects have overlap in their development teams, it makes sense to align their design principles as much as possible. 

This issue will track those developments.

Tagging @Bisaloo.

joshwlambert:
_Question on class standards_: When should a element of a class be required versus optional? 

In {epiparameter} this relates to the list elements of an `<epiparameter>` object. Currently all must be present for the object to be considered valid. But it may be the case that the less important elements (e.g. `$notes`) could be removed and the object is still considered valid.

More generally this question could be applied to a class based on a list, or another non-atomic object, or an atomic object using attributes.

joshwlambert:
From the recent {contactmatrix} dev day I become a bit more familiar with the codebase. The two packages are well aligned from a design perspective. One remaining minor difference is the class name for a dedicated class for a list of objects, `<multi_epiparameter>` and `<contactmatrix_list>`. 

Is it worth renaming one of these so they are consistent, e.g. `<epiparameter_list>` or `<multi_contactmatrix>`. This would not have any functional different.

@Bisaloo what do you think?

Bisaloo:
It makes sense from a conceptual point of view but I'd only do it if it's an easy lift.

I'm still slightly unsure which class name makes more sense :thinking: 

- 189 results for `_list`: https://github.com/search?q=org%3Acran+%2Fclass%5C%28%5Cw%2B%5C%29%5Cs*%28%3C-%7C%3D%29%5Cs*.*%5B%22%27%5D.*_list%5B%22%27%5D%2F&type=code
- 15 results for `multi_`: https://github.com/search?q=org%3Acran+%2Fclass%5C%28%5Cw%2B%5C%29%5Cs*%28%3C-%7C%3D%29%5Cs*.*%5B%22%27%5Dmulti_.*%5B%22%27%5D%2F&type=code

(Note that this misses cases where class is not set by `class<-()`)

Issue #361, opened 2024-07-17
Change from `validate_*()` to `assert_*()`


joshwlambert:

joshwlambert:
Instead of moving to a `validate()` generic. I'll replace the current `validate_*()` naming convention with `assert_*()`. This matches both the naming convention of {checkmate} and of {contactmatrix} (relates #364).

PR #366, opened 2024-08-13
Replace `validate_epiparameter()` with `assert_epiparameter()`

joshwlambert:
***Breaking change***

This PR addresses #361 by replacing the `validate_epiparameter()` function with `assert_epiparameter()`, and adds the `test_epiparameter()` function. The `assert_epiparameter()` and `test_epiparameter()` have the same logic but differ in their output and side-effects. `assert_*()` errors when the object is not an `<epiparameter>` whereas `test_*()` will return `FALSE`. This follows the {checkmate} naming convention and is a development following #364. 

Functions that previously called `validate_epiparameter()` now call `assert_epiparameter()`. 

A bullet point is also added to the `design_principles.Rmd` vignette to give context to the development and naming of the class helper functions and methods.


joshwlambert:
The `test-coverage` GHA is failing, but the tests pass on the `R-CMD-check` across all OSs, so I'll merge this PR and see if the tests continue to fail.

PR #367, opened 2024-08-13
Set up infrastructure to translate print() methods

### LINKED ISSUE #116: Make `print()` method available in other languages
#85 introduced a custom `print()` method for `epiparam` and `epidist` objects. These `print()` methods include text, which could be translated for non-English speaking users.

Bisaloo:
Fix #116.

We still need native speakers to do the actual translation, as documented in https://michaelchirico.github.io/potools/articles/translators.html but this sets up the necessary infrastructure.


DESCRIPTION
```diff
@@ -50,6 +50,7 @@ VignetteBuilder:
 Config/Needs/check: mrc-ide/epireview
 Config/Needs/website: epiverse-trace/epiversetheme, mrc-ide/epireview
 Config/testthat/edition: 3
+Config/potools/style: explicit
```
joshwlambert, line 53: It seems from {potools} documentation that `explicit` is the preferred style. Could you clarify why you chose this over `base`?

joshwlambert (approved):
All looks good. Thanks @Bisaloo for adding this.

It would be good to discuss what the process should be around trying to add translations to the packages. I.e. should we add a line to the README about accepting translation contributions, should we be aiming to add translations for a core set of languages within Epiverse (e.g. es, fr)?

joshwlambert:
@Bisaloo this PR is still a draft, but from what I can see it is ready to be merged unless we want to include any translations within this PR. Can we set the PR as ready to review so it can be merged?

DESCRIPTION
Bisaloo: I wanted fine control over which strings are translatable at this stage. As you can see, I have targeted only the strings for the print method, but not the messages. `explicit` seems like the logical solution to have this level of control.

Bisaloo:
> It would be good to discuss what the process should be around trying to add translations to the packages. I.e. should we add a line to the README about accepting translation contributions, should we be aiming to add translations for a core set of languages within Epiverse (e.g. es, fr)?

Could you share this on the #community channel on slack please? I believe it needs to involve a larger group.

> @Bisaloo this PR is still a draft, but from what I can see it is ready to be merged unless we want to include any translations within this PR. Can we set the PR as ready to review so it can be merged?

Sounds good to me. It was marked as draft to insist on the fact it was low pressure and could easily be delayed/closed if it didn't seem like the best solution.

DESCRIPTION
joshwlambert: Thanks

joshwlambert:
The `test-coverage` workflow is failing and has been sporadically failing since #366. It doesn't relate to this PR so I will go ahead and merge it, and then figure out a solution to apply in a separate PR.

PR #373, opened 2024-08-29
Epireview vignette updates

CarmenTamayo:
PR with updates to the Epireview vignette from Hackathon with Imperial, that mainly focus on updating the process of filtering and subsetting entries thanks to the addition of the new column "parameter_type_short"


joshwlambert (approved):
Thanks for this contribution @CarmenTamayo. I've made a few small tweaks on top of the changes you've made to the vignette. 

I think it's ready to be merged.

PR #379, opened 2024-09-20
Move dispersion from parameter to summary statistic in DB and add dispersion to `create_summary_stats()`

joshwlambert:
This PR redefines the dispersion parameter as a summary statistic for probability distributions that do not use the dispersion parameter in usual parameterisations (e.g lognormal distributions). It is kept as a parameter for distributions that use dispersion as a standard parameter (e.g. negative binomial). This change is made to the parameter database (`parameters.json`). 

The JSON schema (`data_dictionary.json`) is enhanced by adding a `propertyName` `enum` to make sure that summary statistics stored in the database are valid from a set of options. 

The `create_summary_stats()` function is updated to include `dispersion` and dispersion uncertainty as arguments and in the output list. The function documentation is also updated.

### DIFF
--- a/R/epiparameter-utils.R
+++ b/R/epiparameter-utils.R
@@ -226,18 +226,31 @@ create_region <- function(continent = NA_character_,
 #' @param mean_ci_limits A `numeric` vector of length two of the confidence
 #' interval around the mean.
 #' @param mean_ci A `numeric` specifying the confidence interval width,
-#' e.g. 95 would be the 95% CI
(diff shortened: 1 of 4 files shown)


joshwlambert:
Merging this PR even though one of the unit tests is failing as this will be fixed in an upcoming PR. 

Other GHA workflows have also failed since rebasing this branch but this is due to an issue with the workflow and independent of any changes made in this PR.

PR #380, opened 2024-09-23
Move mean to parameter for negative binomial in DB

joshwlambert:
This PR moves the mean and any uncertainty around the mean from the summary statistics object to the parameters object in the JSON database (`parameters.json`). This is to consolidate what is considered a parameter or summary statistic in the package (matching `is_epiparameter_params()`), see also #379.

The JSON schema is updated to explicitly check if the mean, standard deviation and dispersion are being specified correctly depending on the probability distribution. For normal and negative binomial distributions, these should be specified as parameters, whereas for other distributions, e.g. lognormal, they should be specified as summary statistics.

### DIFF
--- a/.github/workflows/validate-json.yaml
+++ b/.github/workflows/validate-json.yaml
@@ -41,7 +41,7 @@ jobs:
             engine = "ajv",
             verbose = TRUE,
             greedy = TRUE,
-            strict = TRUE,
+            strict = FALSE,
             error = TRUE
           )
         shell: Rscript {0}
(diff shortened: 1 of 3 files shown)


joshwlambert:
The `validate-json` workflow is updated to turn off strict schema checking (`strict = FALSE`). This is due to the distribution specific checking requiring parameters (with `"required":`) but the definition of the parameter being in a different scope. See https://github.com/ajv-validator/ajv/issues/1571.

joshwlambert:
The R-CMD-check is not passing on the PR due to failing a unit test checking for a match between the epiparameter library stored in the package (`sysdata`) and regenerating the library from the JSON database. This is part of ongoing work in other branches so I will merge this now and then fix the check in an upcoming PR when other changes have also been merged.

Issue #362, opened 2024-07-29
read an `epiparameter::parameter_tbl()` output to get an `epidist` object


avallecam:
Would it be feasible to make an `epiparameter::parameter_tbl()` output suitable for accessing parameters via `epidist_db()`? This would be ideal for entries that do not fit the `single_epidist = TRUE` statement.

If we want to iteratively access parameters as in this reprex at https://github.com/epiverse-trace/howto/issues/36 we still need to extract the author name as a string to make this work.

``` r
library(tidyverse)
library(epiparameter)

# get set of epidist objects
epidist_input <- epiparameter::epidist_db(
  disease = "covid"
)
#> Returning 27 results that match the criteria (22 are parameterised). 
#> Use subset to filter by entry variables or single_epidist to return a single entry. 
#> To retrieve the citation for each use the 'get_citation' function

# table output
epidist_dataframe <- epidist_input %>% epiparameter::parameter_tbl() %>% 
  slice(8)

epidist_dataframe %>% glimpse()
#> Rows: 1
#> Columns: 7
#> $ disease           <chr> "COVID-19"
#> $ pathogen          <chr> "SARS-CoV-2"
#> $ epi_distribution  <chr> "serial interval"
#> $ prob_distribution <chr> "norm"
#> $ author            <chr> "Yang et al."
#> $ year              <dbl> 2020
#> $ sample_size       <dbl> 131

# convert data frame to epidist
epidist_db(
  disease = epidist_dataframe$disease,
  epi_dist = epidist_dataframe$epi_distribution,
  author = "Yang",
  subset = sample_size == 131
)
#> Returning 1 results that match the criteria (1 are parameterised). 
#> Use subset to filter by entry variables or single_epidist to return a single entry. 
#> To retrieve the citation for each use the 'get_citation' function
#> Disease: COVID-19
#> Pathogen: SARS-CoV-2
#> Epi Distribution: serial interval
#> Study: Yang L, Dai J, Zhao J, Wang Y, Deng P, Wang J (2020). "Estimation of
#> incubation period and serial interval of COVID-19: analysis of 178
#> cases and 131 transmission chains in Hubei province, China."
#> _Epidemiology and Infection_. doi:10.1017/S0950268820001338
#> <https://doi.org/10.1017/S0950268820001338>.
#> Distribution: norm
#> Parameters:
#>   mean: 4.600
#>   sd: 4.400

as_epidist(epidist_dataframe)
[...]

PR #394, opened 2024-10-09
Full package review for v0.3.0

joshwlambert:
This PR is to provide a platform to review the entirety of the package.

Once this review concludes I will release v0.3.0 on GitHub and submit to CRAN.

Please see the `NEWS.md` file for an overview of changes between v0.2.0 and v0.3.0. If you would prefer to review with a partial package review only showing the changes between v0.2.0 and v0.3.0 please let me know and I can open one.

This PR is unconventional as it is not intended for merging or for additional commits (unless minor) and instead comments will be converted to issues and these will be addressed in their own PRs.


R/epiparameter.R
```diff
@@ -0,0 +1,996 @@
...
+      )
+      prob_dist_params <- prob_dist_params[-idx]
+
+      # trunc dist family is truncated so get prob dist by unclassing dist and
+      # extracting name
+      list_dist <- unclass(x$prob_distribution)
+      prob_dist <- gsub(
+        pattern = "dist_",
+        replacement = "",
+        x = class(list_dist[[1]][[1]])[1],
```
chartgerink, line 691: Might it be helpful to add a comment explaining the indexing here? This is convoluted to read and some assistance would help (me in any case).

R/epiparameter.R
```diff
@@ -0,0 +1,996 @@
...
+#' family(ep)
+family.epiparameter <- function(object, ...) {
+  if (inherits(object$prob_distribution, "distcrete")) {
+    prob_dist <- object$prob_distribution$name
+  } else if (inherits(object$prob_distribution, "distribution")) {
+    if (is_truncated(object)) {
+      prob_dist <- gsub(
+        pattern = "dist_",
+        replacement = "",
+        x = class(unclass(unclass(object$prob_distribution)[[1]])[[1]])[1],
```
chartgerink, line 767: Similar to the previous comment on the complicated indexing, might benefit from a comment.

R/epiparameter_db.R
```diff
@@ -0,0 +1,765 @@
...
+#' @inherit epiparameter_db return
+#' @export
+#' @keywords internal
+epidist_db <- function(disease = "all",
+                       pathogen = "all",
+                       epi_name = "all",
+                       author = NULL,
+                       subset = NULL,
+                       single_epiparameter = FALSE) {
+  lifecycle::deprecate_warn(
```
chartgerink, line 265: What version do you want to remove this by? Helpful to note that the deprecation will be completed by a certain version so that you don't get stuck keeping a deprecated function for a long time. Also helps people know by when they need to have moved on.

Good practice on deprecating 👍

R/epiparameter_db.R
```diff
@@ -0,0 +1,765 @@
...
+    l
+  }
+  x <- lapply(x, null2na)
+
+  # format parameters and parameter uncertainty
+  params_uncertainty <- .format_params(x)
+  params <- params_uncertainty$params
+  uncertainty <- params_uncertainty$uncertainty
+
+  # TODO: remove this and move precision from parameter to summary statistic
```
chartgerink, line 562: Is this todo dependent on something external or should this instead be moved to a GitHub issue to track? 

These kinds of comments are easy to lose track of so I would encourage to put it in an issue, fix it, or remove it altogether.

R/epiparameter_db.R
```diff
@@ -0,0 +1,765 @@
...
+      eval(expr = condition)
+    } else {
+      FALSE
+    }
+  } else {
+    # apply condition to <epiparameter> name matching elements
+    lapply(lst, function(x) {
+      if (nse_subject %in% names(x)) {
+        # <epiparameter> is only nested once so no need for recursive search
+        eval(expr = condition, envir = x)
```
chartgerink, line 745: Feels slightly vulnerable to evaluate an expression without adding some checks to it. It is an opportunity for some code to be injected unknowingly. If you can add a check to ensure it's (semi) in the format you would expect, that would make the function more robust as well.

po/R-epiparameter.pot
chartgerink, line 1: I'm unfamiliar with this kind of file. Would you mind expanding on its presence and functionality? It seems harmless,   and I would like to understand better 😊

chartgerink (approved):
Thanks @joshwlambert for opening the full package review! It's really good work and a big achievement 🙌 

Two thumbs up animated GIF

I honestly do not have much to remark. I left some comments for your consideration, but nothing major. I am sure I missed minor issues that can still be improved, but it looks good overall. It is a lot of code so I am hoping for some more eyes on this because I'll inevitably miss aspects (I too, get tired 😄 ). 

I will leave you with two more general questions:

1. Is there a general testing strategy that you applied? It is a lot right now and it would help me understand whether there are any additional strategic considerations to make. For example, did you ensure for every positive (success) test you also had a negative (failure) test to ensure it is not giving a false negative?
2. You know you may get comments on the use of the for loops, instead of vectorizing 😄  I honestly do not mind, but to preempt this question, could you expand on the use of for loops instead of vectorizing in some places?

R/accessors.R
```diff
@@ -0,0 +1,171 @@
...
+#' ep <- epiparameter_db(
+#'   disease = "COVID-19",
+#'   epi_name = "serial interval",
+#'   single_epiparameter = TRUE
+#' )
+#' get_parameters(ep)
+get_parameters.epiparameter <- function(x, ...) {
+  chkDots(...)
+  # extract parameters depending on prob distribution class
+  if (inherits(x$prob_distribution, "distcrete")) {
```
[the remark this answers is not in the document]
joshwlambert: Yes, this is to check the object is `<distcrete>` from the {distcrete} package.

R/epiparameter.R
```diff
@@ -0,0 +1,996 @@
...
+}
+
+#' Test whether an object is a valid `<epiparameter>` object
+#'
+#' @param x An \R object.
+#'
+#' @return A boolean `logical` whether the object is a valid `<epiparameter>`
+#' object.
+#' @export
+test_epiparameter <- function(x) { # nolint cyclocomp_linter
```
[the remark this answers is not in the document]
joshwlambert: I like having the `lintr` check the cyclomatic complexity as I feel it does help me not to write overly complex functions. The default is 15 (https://lintr.r-lib.org/reference/cyclocomp_linter.html) which I think I'll leave the same for now. As you can see I don't comply with the cyclocomp linter too strictly by occasionally using `# nolint cyclocomp_linter`.

I think the best approach would either be to comply with the linter more strictly and remove the `# nolint` flags or use them sparingly like the current approach. I think all other Epiverse-TRACE packages use the default linter so it would be good to try and stay consistent across the organisation.

R/epiparameter.R
joshwlambert: This has been resolved in PR #398. This PR generalises `.get_mixture_dist()` to `.distributional_family()` which is now called here instead of the multiple consecutive subsetting.

R/epiparameter.R
joshwlambert: Thanks both for this suggestion. I've made improvements to these parts of the code in PR #398 by generalising `.get_mixture_dist()` to `.distributional_family()` which enables the access of the distribution names, for both transformed (e.g. mixture or truncated) or untransformed (e.g. gamma or lognormal) distributions. The `.distributional_family()` took inspiration from some code posted on mitchelloharawild/distributional#110 and as stated on that issue if a new function gets added to {distributional} we can potentially remove some code from this package and rely on their exported function.

R/epiparameter_db.R
joshwlambert: I'll remove this function in v0.5.0. I've added a details note to `lifecycle::deprecate_warn()` in PR #399 to tell users about when the function will be removed. In {epiparameter} v0.4.0 I will change `lifecycle::deprecate_warn()` to `lifecycle::deprecate_stop()`.

R/epiparameter_db.R
joshwlambert: Thanks, I've logged in #400 and will resolve in the next version and will remove this `# TODO` comment.

po/R-epiparameter.pot
joshwlambert: This is a file for translating messages that are generated by {epiparameter} functions. It was added in PR #367. See the [{potools}](https://michaelchirico.github.io/potools/) package for more info. There currently are not any translations included in the package, but the infrastructure is here so now we just need the equivalent messages in other languages.

R/epiparameter.R
```diff
@@ -0,0 +1,996 @@
...
+        "statistics.\n Probability distribution is assumed not to be ",
+        "discretised or truncated."
+      )
+    }
+    if (!inherits(prob_distribution, c("distribution", "distcrete"))) {
+      message("Unparameterised <epiparameter> object")
+    }
+  }
+
+  if (epi_name == "offspring_distribution") {
```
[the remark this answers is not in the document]
joshwlambert: I have made some improvements to this in PR #401 to try and match non-delay distributions, which currently include offspring distributions and case fatality risks. 

The data dictionary used in the JSON validation workflow uses an `enum` field to check that the `epi_name` is within a predefined set https://github.com/epiverse-trace/epiparameter/blob/main/inst/extdata/data_dictionary.json#L23.

However, this is a temporary solution and needs to be improved to be more scalable as more non-delay distributions are added to the package, and to accommodate users creating `<epiparameter>` objects with a variety of parameter names. We could impose that the `epi_name` argument must also be within the same set as defined in the data dictionary in when specified by the user in `epiparameter()` but I don't think there would be that much benefit, while it could hinder users and put more development burden adding more parameter types. 

Happy to discuss this further to find a more optimal solution. This can optionally be filed as an issue to continue discussion after the package review.

R/epiparameter.R
```diff
@@ -0,0 +1,996 @@
...
+#'   ),
+#'   method_assess = create_method_assess(
+#'     censored = TRUE
+#'   ),
+#'   notes = "No notes"
+#' )
+epiparameter <- function(disease,
+                         pathogen = NA_character_,
+                         epi_name,
+                         prob_distribution = create_prob_distribution(
```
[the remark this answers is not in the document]
joshwlambert: I personally prefer to use the helper functions as argument defaults as I think it helps navigate users towards these functions which they may otherwise not know exist. I agree in this case it is an unnecessary abstraction, but for consistency with the rest of the package I will leave as is.

R/epiparameter.R
```diff
@@ -0,0 +1,996 @@
...
+  stopifnot(
+    "uncertainty must be provided for each parameter" =
+      length(param_names) == length(epiparameter$uncertainty),
+    "parameters and uncertainty must be named and match" =
+      identical(param_names, names(epiparameter$uncertainty)) ||
+      is.na(param_names)
+  )
+
+  # call epiparameter validator
+  assert_epiparameter(epiparameter)
```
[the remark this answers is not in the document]
joshwlambert: I agree this is superfluous. I like having it to ensure the object created is valid. I think it is impossible to get to the `assert_epiparameter()` call with an invalid `<epiparameter>` object, but just as an extra layer of defence I've added it. 

As the time to run this extra function is negligible I do not think there is much reason to remove, but if users were creating many thousand/million `<epiparameter>` objects this would be a good first thing to remove for a speed gain.

If there are other reasons to remove that I've overlooked please let me know.

R/epiparameter.R
```diff
@@ -0,0 +1,996 @@
...
+
+#' Assert an object is a valid `<epiparameter>` object
+#'
+#' @param x An \R object.
+#'
+#' @return Invisibly returns an `<epiparameter>`. Called for side-effects
+#' (errors when invalid `<epiparameter>` object is provided).
+#'
+#' @export
+assert_epiparameter <- function(x) {
```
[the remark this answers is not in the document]
joshwlambert: Thanks for raising this. I've managed to refactor these functions in PR #402 which has reduced the duplication and enhanced the messages/errors presented to users.

R/epiparameter_db.R
joshwlambert: I like this suggestion, but nothing comes immediately to mind as to how we could test this conditions before they get evaluated. Do you have ideas for how to do this?

R/epiparameter.R
```diff
@@ -0,0 +1,996 @@
...
+  if (isFALSE(is_parameterised(x))) {
+    stop("<epiparameter> is unparameterised", call. = FALSE)
+  }
+  unlist <- length(x$prob_distribution) == 1
+  if (inherits(x$prob_distribution, "distcrete")) {
+    out <- x$prob_distribution$d(at)
+  } else {
+    out <- stats::density(x$prob_distribution, at = at)
+  }
+  out <- if (unlist) unlist(out, recursive = FALSE) else out
```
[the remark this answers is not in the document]
joshwlambert: This change has been implemented for `density.epiparameter()`, `cdf.epiparameter()` and `quantile.epiparameter()` in PR #403. 

In that PR I've also changed the condition from `length(x$prob_distribution) == 1` to `is.atomic(out)`. I couldn't remember if there was a specific reason why I used the length of the object previously but given that the unlisting is to ensure an atomic vector is returned I thought `is.atomic()` would be more readable. All tests pass so hoping that the conditions are functionally equivalent.

joshwlambert:
Thank you both for helpful comments and suggestions @chartgerink & @TimTaylor! I've made several PRs with improvements resulting from this review which are either linked in my responses, or linked at the bottom of this PR (or both). 

In response to some other questions and comments:


> 1. Is there a general testing strategy that you applied? It is a lot right now and it would help me understand whether there are any additional strategic considerations to make. For example, did you ensure for every positive (success) test you also had a negative (failure) test to ensure it is not giving a false negative?

There currently is not a testing strategy. I usually try and write unit tests for both exported and internal functions for expected behaviour under most common usage scenarios, and then a couple of failure expectations to check the function errors nicely. 

Given the namespace of this package is quite large (relative to other Epiverse-TRACE package), I've thought about moving to only testing exported functions. Some tests have been added to check that issues are resolved correctly, so it's been a fairly organic process of adding tests over time. I'd be happy to discuss possible testing strategies to make the process a bit more formal. 

> 2. You know you may get comments on the use of the for loops, instead of vectorizing 😄 I honestly do not mind, but to preempt this question, could you expand on the use of for loops instead of vectorizing in some places?

There is often not a formal logic on when to use vectorisation versus loops. Usually when I write code I will write loops (that's just how my brain works). Then I'll usually see how the code works once drafted and see if it can be optimised, usually by vectorising loops. Sometimes I've found I'm not able to vectorise so the loop remains, others I've found the loop more readable so left as is. 

If there are specific functions that contain either loops or vectorised statements that you think could be converted to the other, please raise them as an issue and I'd be happy to update (e.g, https://github.com/epiverse-trace/simulist/pull/150).

> At present `new_epiparameter()` is only used within `epiparameter()`. Is it worth pulling the functionality in to `epiparameter()` itself. A minimal constructor is normally useful when you need to quickly recreate an object after a method has dropped it's class.

Currently, the split in functionality is `epiparameter()` does the input checking and handles the parameter uncertainty after the `<epiparameter>` object has been constructed in `new_epiparameter()`. `new_epiparameter()` calculates the distribution parameters from summary statistics, if available. I could potentially merge them as, like you say, `new_epiparameter()` is only ever called within `epiparameter()`. I'd be interested to hear your opinions on when to have a minimal class constructor and whether this should be strictly applying the class attribute. I saw you had a similar issue in {incidence2} (reconverse/incidence2#103) so would be good to get your thoughts as I assume a similar design choice could be taken across packages.

> `chkDots()` seems to be used in some methods but not others.

`chkDots()` should only be used in method where extra arguments should not be passed via `...`. I will check if they are being used consistently and whether any other methods should call it.

> I see you use `assert_epiparameter()` at the start of some epiparameter methods in other files (e.g. `convert_params_to_summary_stats.epiparameter()`). Is this a "belts and braces" approach or do you think there is something fragile about the epiparameter class that makes it likely to be broken by users?

Yes on the latter. It is incase an `<epiparameter>` is dispatched but the object has been invalidated by a user between being imported/constructed and the generic being called. I currently haven't implemented any subsetting or assignment operator methods for `<epiparameter>` (e.g. `$`, `[`, etc.) so this is to offset that. 

> As discussed offline, I wonder if there is an intermediate more tidy-like tabular data structure that would be useful. I'm not sure but leaving here for my own pondering ...

This relates to #362, I will copy over the code chunk to that issue to continue discussion there. I won't make any changes with regard to this point before the release, but can give it some thought over the next development cycle.

PR #404, opened 2024-10-22
Decrease R mininum version to 4.0.0

joshwlambert:
This PR changes the minimum R version required by {epiparameter} to 4.0.0, after it was incremented to 4.1.0 in PR #384 (due to mitchelloharawild/distributional#128). 

This PR also adds a R-CMD-check CI run for R 4.0 to check that the R package passes the check on the version stated in the `DESCRIPTION`. 

This PR initially checked to see if the minimum R version could be reverted to 3.6.0 as it was before #384, however, due to {epireview} depending on {flextable} which in turn depends on {gdtools} which requires R 4.0, then this is required for the R-CMD-check workflow to install {epireview}.

Usage of the base pipe (`|>`) has been replaced in a vignette by the pipe from {magrittr} (`%>%`), with {magrittr} added as a suggested dependency. 

The note in the `design_principles.Rmd` vignette on breaking the Epiverse R version policy has been removed as this will not longer be the case.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -59,6 +59,7 @@ jobs:
           - {os: ubuntu-latest,   r: 'devel', http-user-agent: 'release'}
           - {os: ubuntu-latest,   r: 'release'}
           - {os: ubuntu-latest,   r: 'oldrel-1'}
+          - {os: ubuntu-latest,   r: '3.6'}
 
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


joshwlambert:
This PR fails the R-CMD-check CI for both R v3.6.0 and v4.0.0. Due to {epireview} depending on {gdtools} which requires R 4.0 and then {distributional} which released v0.5.0 using the base pipe (`|>`). The base pipe has since been removed from {distributional} but this has not yet been released to CRAN.

Therefore, I will close this PR without merging, and retain the minimum R version of 4.1.0 for {epiparameter} and add a explicit R-CMD-check CI for R 4.1 on a separate PR. 

The alternative would be to wait to for a new {distributional} release to enable the minimum R version to be 4.0, but this seems like a marginal gain and would be a blocker to releasing {epiparameter}.

joshwlambert:
Closing, superseded by #405.

PR #442, opened 2025-02-04
Add offset to `<epiparameter>`, class methods and helper functions

joshwlambert:
This PR adds the functionality for an `<epiparameter>` object to have a distribution offset. See epiverse-trace/epiparameterDB#24 for addition of `offset` parameters to the database.

An `offset` argument is added to `create_prob_distribution()`, with the `offset` being stored as a class attribute of the returned object. This avoids the requirement of the distribution object to handle offsets explicitly (for example {distributional} and {distcrete} currently do not handle offsets). 

The `<epiparameter>` distribution functions (`density()`, `cdf()`, `quantile()` and `generate()`) are updated to use the offset. By default, a zero offset will produce the same behaviour as previous versions.

The `<epiparameter>` `format()` and `print()` methods are updated to print the `offset` if it is non-zero.

Two unit tests are added that check `create_prob_distribution()`.

This PR also temporarily adds the dependency on {epiparameterDB} from the ~`@offset` branch~ `main` branch on GitHub. ~The dependency will be removed from a specific branch when the Mpox parameters are merged into `main` in {epiparameterDB}~, the remote dependency will be removed once the new version of {epiparameterDB} is on CRAN.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -56,6 +56,8 @@ Suggests:
     testthat (>= 3.0.0),
     vdiffr (>= 1.0.7),
     visNetwork
+Remotes:
+    epiverse-trace/epiparameterDB@offset
 VignetteBuilder: 
     knitr
 Config/Needs/check: mrc-ide/epireview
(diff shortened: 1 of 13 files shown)


joshwlambert:
Snaphots have been updated due to changes upstream in {epiparameterDB} (`main` branch on GitHub).

PR #446, opened 2025-04-15
Test GitHub actions for R-devel

joshwlambert:
This PR is just to test that {epiparameter} passes GHA workflows due to it potentially causing an issue on epiverse-trace/epiverse#7. R-devel is >4.5.0.


joshwlambert:
This PR passes R CMD check on all OSs and R versions. 

The issue on epiverse-trace/epiverse#7 was due to tests running in parallel on R-devel and unrelated to {epiparameter}.

PR #453, opened 2025-08-07
Fix suppressMessages() function signature

Bisaloo:
Detected with the experimental https://github.com/Bisaloo/signaturechecker package.

It's a fun coincidence this managed to go undetected:

- `suppressMessages()` did warn that `regexp=` is not a valid argument
- but it is wrapped in `expect_error()`, with no `regexp=` value, so it's silently ignored.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -66,4 +66,4 @@ Encoding: UTF-8
 Language: en-GB
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.3.2
+RoxygenNote: 7.3.3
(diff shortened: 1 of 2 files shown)


joshwlambert:
Thanks @Bisaloo for the PR. The reason the unit test was failing after the `suppressMessage()` signature was changed was due to the `epiparameter()` function not correctly calling `create_prob_distribution()` and the `prob_distribution_params` argument being silently swallowed by `...`. This is due to this function being missed in the update when `epiparameter()` was changed to use `create_prob_distribution()`. I've fixed this in 2cbdeafad67bd18b06e7238daec945761e5fa881.

joshwlambert (approved):

PR #486, opened 2026-06-02
Remove deprecated `epidist_db()`

joshwlambert:
This PR removes the `epidist_db()` function which was deprecated in v0.3.0 (renamed to `epiparameter_db()`) with a notice that it would be removed in v0.5.0 (upcoming release). 

The dependency on the {lifecycle} package is removed as it was only used for `epidist_db()` deprecation, as well as removing the `man/figures/lifecycle-*.svg` badge files.

The [ROR](https://ror.org/) is added to the `epiparameter-package.Rd` file.


joshwlambert:
Much of the {lifecycle} related content that is removed in this PR was added in PR #360.

